PR #2, opened 2015-05-13
dynamically determine python flags for cfg

novocaine:
* in build.rs, call the python interpreter located by pkgconfig
and use sysconfig to determine the flags it was built with.
* pass these flags to the build via --cfg
* refactor existing references to features to refer to these
instead

background - I'm trying to get this to build on os x, and apparently the homebrew python 2.7 is a utf-16 build (?!) so seemed logical to get this working (another pr coming soon for some blips in string conversion under utf-16).

this would need to be replicated across to python 3 also; would be keen to hear your thoughts on how best to share code between the two build.rs scripts for the two ffis.

### DIFF
--- a/python27-sys/Cargo.toml
+++ b/python27-sys/Cargo.toml
@@ -21,21 +21,11 @@ exclude = [
 [dependencies]
 libc = "*"
 
-[build-dependencies]
-pkg-config = "0.3"
+# [build-dependencies]
+# pkg-config = "0.3"
 
-[features]
-# Features correspond to possible settings used when compiling python.
-# It's probably a misuse to use cargo features for this;
-# ideally we'd auto-detect these in the build script.
-default = ["Py_USING_UNICODE", "Py_UNICODE_WIDE", "WITH_THREAD"]
-
-Py_USING_UNICODE = []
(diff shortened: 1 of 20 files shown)


dgrunwald:
Nice! I planned to do something like this, but I didn't realize the values were available through python `sysconfig`.

Yeah, I'm pretty unsure about how to share code between python2.7 / python3 at the moment.
For python 3, I'm using the PEP 384 stable ABI, so it doesn't depend on that many build options. I don't even know PEP 384 is supported without WITH_THREAD or with Py_TRACE_REFS (I think I've seen some comments saying that it doesn't support Py_DEBUG), so we might not need this logic there (unless python3-sys gets extended to offer more than just the PEP 384 API).

For rust-cpython, I'd really like to have one codebase that uses #[cfg] to build against python27-sys or python3-sys. (although on crates.io, cpython27 and cpython3 might still have to be separate packages?).

python27-sys / python3-sys could be different branches of a single repository, but on the other hand, I like to have them directly in the rust-cpython repository (and I dislike git submodules).
The easiest solution is probably just to duplicate the build.rs and share code by manually merging between python27-sys and python3-sys -- I expect changes to python27-sys will be relatively rare once it's working on more than just my machine :)

PR #3, opened 2015-05-17
fix issues in bb13ec, support utf16, python fom PATH

novocaine:
* fix ucs4 build broken by bb13ec
* add utf16 decoding to unicode.from_py_object for
  narrow unicode builds
* change unicode narrow/wide cfg flag to be
  Py_UNICODE_SIZE_4 not Py_UNICODE_WIDE, which doesn't
  appear in sysconfig
* support framework builds on os x
* python27-sys exports compilation flags as cargo vars,
  and rust-python resurrects them as cfg flags
* travis runs against local python27-sys
* rust-cpython depends on git python27-sys, because
  the one on cargo is now incompatible with it (since bb13ec)

Okay, so things were a little complicated than I previous thought :/

1. The CI was actually running off crates.io's python27-sys, so I didn't realise I had a bunch of things wrong and had actually broken the build - sorry :/. I am still learning about the rust build system. In this commit, I've altered the travis.yml to force tests to run against a local python27-sys. Maybe there's a better way than this? But I think the CI needs to fully reflect the code in the repo to avoid this happening again.

2. --cfg flags don't actually propagate to dependent crates, there's a separate system for exporting vars in cargo. So previously rust-python wasn't seeing cfg flags it needed in order to compile, and was broken. python27-sys now exports the python flags, and rust-python picks them up and resurrects them as the same cfg flags - so we can still use identical syntax in the two packages.

3. Fixed utf-16 decoding, added explicit os x fixes, and can now run off the python in your PATH, which works great with virtualenv and is how I managed to flip between testing against wide/narrow python builds.

If you like, I'm happy to split this into the commits which fix stuff I broke, and the commits which add new features for os x/utf16.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: rust
 env:
   global:
     - secure: g4kCg8twONwKPquuJmYrvGjo2n0lNtWTbyzFOITNn8FgCxNK2j38Qc9/UhErTR3g3rDjVzsTHZ8FTH7TJZrOK1Nzz90tJG6JHqUv77ufkcBlxgwwjilOz84uQhkDTMpLitMEeQDLEynKeWbxrjtc5LIpjEkxOPk5eiqwzKRN14c=
+before_script:
+  - mkdir -p .cargo && echo 'paths = ["python27-sys"]' > .cargo/config
 script:
   - cargo build --verbose
   - cargo test --verbose
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,8 +18,12 @@ exclude = [
     ".gitignore",
     ".travis.yml",
 ]
+build = "build.rs"
 
 [dependencies]
-python27-sys="*"
 libc="*"
 num="*"
+
+# TODO: crates.io version doesn't have necessary conditional cfg flags yet
+[dependencies.python27-sys]
+git = "https://github.com/dgrunwald/rust-cpython.git"
--- a/python27-sys/Cargo.toml
+++ b/python27-sys/Cargo.toml
@@ -12,6 +12,7 @@ homepage = "https://github.com/dgrunwald/rust-python27-sys"
 repository = "https://github.com/dgrunwald/rust-python27-sys.git"
 license = "Python-2.0"
 authors = ["Daniel Grunwald <daniel@danielgrunwald.de>"]
+links = "python27"
 build = "build.rs"
 exclude = [
     ".gitignore",
--- a/python27-sys/build.rs
+++ b/python27-sys/build.rs
@@ -38,15 +38,7 @@ static SYSCONFIG_VALUES: [&'static str; 1] = [
 /// Examine python's compile flags to pass to cfg by launching
 /// the interpreter and printing variables of interest from 
 /// sysconfig.get_config_vars.
-fn get_config_vars() -> Result<HashMap<String, String>, String>  {
-    let exec_prefix = pkg_config::Config::get_variable(
-        "python-2.7", "exec_prefix").unwrap();
-
-    // assume path to the pkg_config python interpreter is 
-    // {exec_prefix}/bin/python - this might not hold for all platforms, but
-    // the .pc doesn't give us anything else to go on
-    let python_path = format!("{}/bin/python", exec_prefix);
-
+fn get_config_vars(python_path: &String) -> Result<HashMap<String, String>, String>  {
     let mut script = "import sysconfig; \
 config = sysconfig.get_config_vars();".to_owned();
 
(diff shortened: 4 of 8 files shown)


novocaine:
p.s., the ci is breaking after this rev, but that appears to be due to some regression in rustdoc unrelated to what I've done ..!

.travis.yml
```diff
@@ -2,6 +2,8 @@ language: rust
 env:
   global:
     - secure: g4kCg8twONwKPquuJmYrvGjo2n0lNtWTbyzFOITNn8FgCxNK2j38Qc9/UhErTR3g3rDjVzsTHZ8FTH7TJZrOK1Nzz90tJG6JHqUv77ufkcBlxgwwjilOz84uQhkDTMpLitMEeQDLEynKeWbxrjtc5LIpjEkxOPk5eiqwzKRN14c=
+before_script:
+  - mkdir -p .cargo && echo 'paths = ["python27-sys"]' > .cargo/config
```
dgrunwald: Interesting! I didn't know about cargo path overrides.

Is there any reason not to use these when working with the rust-cpython repository? I think we might want to commit `.cargo/config` to version control.

Cargo.toml
```diff
@@ -18,8 +18,12 @@ exclude = [
...
+build = "build.rs"
 
 [dependencies]
-python27-sys="*"
 libc="*"
 num="*"
+
+# TODO: crates.io version doesn't have necessary conditional cfg flags yet
+[dependencies.python27-sys]
+git = "https://github.com/dgrunwald/rust-cpython.git"
```
dgrunwald: Instead of referencing the git repository with `git = ...`, it's also possible to use `path = "python27-sys"` here.
But actually I think the best solution would be to revert to `python27-sys="*"`, and rely on the cargo path override (to be added to version control) during development. Then we just need to make sure to release python27-sys before releasing a new cpython package.

python27-sys/Cargo.toml
```diff
@@ -12,6 +12,7 @@ homepage = "https://github.com/dgrunwald/rust-python27-sys"
 repository = "https://github.com/dgrunwald/rust-python27-sys.git"
 license = "Python-2.0"
 authors = ["Daniel Grunwald <daniel@danielgrunwald.de>"]
+links = "python27"
```
dgrunwald: Why is this necessary? Shouldn't the build script output the linker flags? Especially since there's `python27` and `python2.7` depending on platform...

src/objects/string.rs
```diff
@@ -126,17 +174,7 @@ impl <'p, 's> FromPyObject<'p, 's> for Cow<'s, str> {
...
-                match char::from_u32(c) {
-                    Some(c) => s.push(c),
-                    None => {
-                        let e = try!(exc::UnicodeDecodeError::new(
-                            py, cstr!("utf-32"), u32_as_bytes(u), i .. i+1, cstr!("invalid code point")));
-                        return Err(PyErr::new(e));
-                    }
-                }
-            }
+            let s = unicode_buf_to_str(py, u).unwrap();
```
dgrunwald: This should be `try!()` instead of `unwrap()`

.travis.yml
novocaine: yes we could, I just wasn't sure whether you had that in mind and I wonder if it's impolite to take precedence over any configs further up the dir tree..? precedence rules aren't made very clear here:

http://doc.crates.io/config.html

Cargo.toml
novocaine: okay sure

python27-sys/Cargo.toml
novocaine: Not 100% sure if I understand the question but here's the explanation .. 

I think this is required in order for the cargo: variables that python27-sys exports to be visible to dependents (such as rust-cpython) as environment variables. cargo seems to use this as a prefix when translating the cargo: vars into environment variable names.

http://doc.crates.io/build-script.html#the-links-manifest-key

For example, in rust-cpython/build.rs above you can see I am unpacking DEP_PYTHON27_PYTHON_FLAGS - well, I never put 'dep_python27' in the cargo: exports at the bottom of python27-sys/build.rs - cargo did. So I think we need to put something here. 

(personally, I think that passing info to dependents should be an orthogonal concept to binding to native libraries - but that doesn't seem to be how it is).

src/objects/string.rs
novocaine: yes will fix

python27-sys/Cargo.toml
novocaine: oh, i think I just realised the misunderstanding - this does not control the linker flags at all, its just a declaration that this crate links to the named native lib for the purposes of signalling to dependent crates looking for a provider of it. again see the doc.crates.io link above.

python27-sys/build.rs
```diff
@@ -97,19 +93,143 @@ fn cfg_line_for_var(key: &str, val: &str) -> Option<String> {
...
+    }
+
+    let out = String::from_utf8(out.stdout).unwrap();
+    return Ok(out);
+}
+
+#[cfg(not(target_os="macos"))]
+fn get_rustc_link_lib(enable_shared: bool) -> Result<String, String> {
+    if enable_shared {
+        Ok("cargo:rustc-link-lib=python2.7".to_owned())
```
novocaine: your point about not hardcoding "python2.7" as the link target might apply here, though - I will see if we can extract this from sysconfig.

dgrunwald:
I've merged this and fixed the build by disabling the `cargo doc` invocation.

PR #5, opened 2015-05-23
Python 3 build support

novocaine:
* Add python3-sys to rust-cpython as an optional feature, and
  make python27-sys also optional, but still the default
* Parametrise python27-sys/build.rs so that it is python
  version independent, and clone it into python3-sys/build.rs.
  Hopefully this can continue to be maintained as an identical
  file.
* python27-sys and python3-sys gain features for explicitly
  selecting a python version to link to. for python27-sys,
  there's currently only python27; for python3-sys there's
  python 3.4 and 3.5.
* explicitly tell travis to use nightlies (seems to have
  started trying to use 1.0.0)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: rust
+rust: nightly
 env:
   global:
     - secure: g4kCg8twONwKPquuJmYrvGjo2n0lNtWTbyzFOITNn8FgCxNK2j38Qc9/UhErTR3g3rDjVzsTHZ8FTH7TJZrOK1Nzz90tJG6JHqUv77ufkcBlxgwwjilOz84uQhkDTMpLitMEeQDLEynKeWbxrjtc5LIpjEkxOPk5eiqwzKRN14c=
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,16 @@ build = "build.rs"
 libc="*"
 num="*"
 
+# These features are both optional, but you must pick one to 
+# indicate which python ffi you are trying to bind to.
 [dependencies.python27-sys]
 path="python27-sys"
+optional = true
 
+[dependencies.python3-sys]
+path = "python3-sys"
+optional = true 
+
+[features]
+# Maybe one day python 3 should be the default. But not this day.
+default = ["python27-sys"]
--- a/build.rs
+++ b/build.rs
@@ -1,11 +1,28 @@
 use std::env;
+use std::io::Write;
 
 const CFG_KEY: &'static str = "py_sys_config";
+
+#[cfg(feature="python27-sys")]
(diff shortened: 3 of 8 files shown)


novocaine:
to attempt to build rust-cpython against python3-sys, the command is 

`cargo build --features python3-sys --no-default-features`

obviously this doesn't currently work yet as rust-cpython doesn't address the python 3 ffi.

is supporting python 3 in rust-cpython - i.e. it acts as a version independent layer - a goal of the project?

dgrunwald:
I'd like to support python 3, but I don't think we can be a version independent layer when it comes to `bytes`/`str`/`unicode`. There's going to be some minor API differences depending on the target python version. Maybe we'll even have some differences depending on whether someone wants to target a specific python version like 3.4, or use PEP 384 for binary compatibility with all versions >=3.3: For example, rust-cpython currently allows interpreting a tuple as a `&[PyObject]` slice, but that's only possible if `PyTupleObject` uses a contiguous allocation. So far all python versions do, but it's not guaranteed by PEP 384. (that aside, maybe we should just remove the tuple-as-slice functionality; it's the main reason rust-cpython can't work on Rust 1.0.0)

But as long as the differences due to python versions are minor, it should still be possible for someone to use rust-cpython as a version independent layer by avoiding version-specific functions and e.g. treating the return type of `str::to_py_object()` as "something implementing `PythonObject`" instead of a concrete `PyObject` or `PyString`.

I'm not sure how to handle this scenario with crates.io -- do we provide a single rust-cpython package that can use python 2 or python 3 depending on cargo features, or do we provide separate packages built from the same codebase? (I think currently prefer using cargo features)

dgrunwald:
Looking at the list of compiler errors with python 3:
* many related to `str` vs. `unicode`
* many related to the combination of `int` and `long` in python 3
* some macros not available due to PEP 384 restrictions, (esp. PyList/PyTuple), these can be replaced with the corresponding function calls
* old-style classes were removed from python 3 -- we just need disable these with a `#[cfg]`
* critical: `ffi::PyEval_InitThreads()` is missing due to PEP 384 restrictions. I'll have to look into whether we can safely initialize the python runtime without this; this point might kill the idea of using PEP 384 :(

novocaine:
I think I prefer using cargo features also, and I think that its easier to keep on top of everything if there's a single package in terms of versioning and CI. 

I think it's important to support Python 3. Happy to peck away at some of the smaller python 3 problems. No idea what to do about `PyEval_InitThreads`.

dgrunwald:
`PyEval_InitThreads` isn't actually missing, I just misread which header file it was from. It actually was missing due to the change from `#[cfg(feature = "WITH_THREAD")]` to `#[cfg(py_sys_config = "WITH_THREAD")]` (I already fixed this).

PR #18, opened 2015-07-07
A method for accessing dictionary items.

traff:


dgrunwald:
`PyDict_Items` returns a new reference, not a borrowed pointer. Your code will cause a reference leak.

Also, `PyObject::from_borrowed_ptr` has undefined behavior if `PyDict_Items` fails (at least `MemoryError` is possible).

Use the helper functions in `err` for handling the return value of python FFI calls.
`err::cast_from_owned_ptr_or_panic()` would work here.

dgrunwald:
I've changed the return type of `items()` to `Vec<(PyObject, PyObject)>`, as that's easier to iterate with a Rust for loop.
I've renamed your method `items_list()`, but it's probably not necessary anymore as `dict.items().into_py_object()` produces the same list (even if slightly less efficiently).

PR #20, opened 2015-07-17
Implement ToPyObject for HashMap->dict conversion.

traff:


dgrunwald:
Try to avoid `unsafe` code if possible. The `PyDict` impl already has wrapper functions around these python FFI calls.

Interesting question: what should happen if `PyDict_SetItem` fails (e.g. due to a `__hash__` method raising an exception)?
Unfortunately `ToPyObject` doesn't really have a way of reporting any errors, as the other impls can't fail (except for out-of-memory errors).

traff:
Valid points.
If PyDict_SetItem fails, I'd prefer that it panics.

traff:
Done.

PR #21, opened 2015-07-21
Accessing PyFrameObject and PyCodeObject.

traff:
This is implemented for Python-2.7 only. Allows to access CPython frame and code objects from Rust.


dgrunwald:
PyFrameObject is defined in `frameobject.h`, not `pystate.h` (the latter only has a forward declaration).
Please keep the organization of the rust files identical to the python headers.

PR #23, opened 2015-08-02
Fix function and method name fields

murarth:


murarth:
Previously, attempting to access `py_fn.__name__` would cause a segfault. This can be seen with the example extension, `hello`.

PR #24, opened 2015-08-02
Implement `eval` and `run` to execute Python code

murarth:


dgrunwald:
`PyRun_String` is declared in `pythonrun.h`, not in `eval.h`. Also, it's declared as a macro, not a function.

Please keep the structure of the `python-sys` modules consistent with the python C headers.

murarth:
I've updated the code.

PR #29, opened 2015-08-26
Add tests for list.rs. Still missing tests for extract.

ehiggs:
This PR adds coverage for the list module.

I tried to also put in code to test extraction into vectors, but the following fails to compile:

```rust
    #[test]                                                                     
    fn test_extract() {                                                         
        let gil = Python::acquire_gil();                                        
        let py = gil.python();                                                  
        let v = vec![2, 3, 5, 7];                                               
        let list = v.to_py_object(py).into_object().cast_into::<PyList>().unwrap();
        let v2 = list.into_object().extract::<Vec<i32>>().unwrap();             
        assert_eq!(v, v2);                                                      
    }      
```

The error is:

```
$ cargo test
   Compiling cpython v0.0.4 (file:///home/ehiggs/src/download/rust-cpython)
src/objects/list.rs:251:37: 251:58 error: the requirement `for<'s>  : 's` is not satisfied (`expected bound lifetime parameter 's, found concrete lifetime`) [E0279]
src/objects/list.rs:251         let v2 = list.into_object().extract::<Vec<i32>>().unwrap();
                          
```

I was on IRC and some users suggested relaxing the bound constraints on extract so they are only over `'p` (instead of `for<'s, 'p>`) but I couldn't get it to work.

### DIFF
--- a/src/objects/list.rs
+++ b/src/objects/list.rs
@@ -160,3 +160,86 @@ impl <'python, 'source, 'prepared, T> ExtractPyObject<'python, 'source, 'prepare
     }
 }
 
+#[cfg(test)]
+mod test {
+    use std;
+    use python::{Python, PythonObject};
+    use conversion::ToPyObject;
+    use objects::PyList;
+
+    #[test]
+    fn test_len() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let v = vec![1,2,3,4];
+        let list = v.to_py_object(py).into_object().cast_into::<PyList>().unwrap();
+        assert_eq!(4, list.len());
+    }
+
+    #[test]
+    fn test_get_item() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let v = vec![2, 3, 5, 7];
+        let list = v.to_py_object(py).into_object().cast_into::<PyList>().unwrap();
+        assert_eq!(2, list.get_item(0).extract::<i32>().unwrap());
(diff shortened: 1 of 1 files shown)


dgrunwald:
`.into_object().cast_into::<PyList>().unwrap()` is a no-op, the code can be simplified to `let list = v.to_py_object(py);`

The problem with `impl ExtractPyObject for Vec<T>` is that it requires `T : ExtractPyObject<'python, 's, 'p>` for all `'s` and `'p`, which is impossible because `'python` must outlive `'s`.
I'm not sure if there's a way to quantify a higher-ranked-trait-bound only over the lifetimes smaller or equal to a given lifetime.
Imaginary syntax:
    where T: (for<'s, 'p> ExtractPyObject<'python, 's, 'p> where 's : 'python)

Another option would be to unnecessarily quantify over `'py` instead of using `'python`:
    where T: for<'py:'s, 's, 'p> ExtractPyObject<'py, 's, 'p>

Unfortunately that doesn't seem to have the desired effect.

Here's a slightly reduced test case for the problem: http://is.gd/bl0rGA

Note that the reason the lifetimes are so complicated is that we support extracting non-owned types like `&str` or `&[T]` as part of the `py_fn!()`/`py_argparse!()` macros. In the case of `PyList`, the higher-ranked lifetime bound is necessary for both `'source` and `'prepared`, because both refer to local variables in `fn extract()`. Getting items from a list has to clone the `PyObject` because any python operation (such as extracting one of the list elements) could potentially mutate the list.

PR #30, opened 2015-08-26
Add support for the Sequence protocol

ehiggs:
This adds support for the Sequence protocol.

Some notes: 

I use `into_object().cast_into::<PySequence>().unwrap()` in the tests because otherwise `to_py_object` would just resolve to the base type (`PyList`, `PyTuple`, etc).

I can't `extract` which I think gets in the way of making interesting assertions on `get_slice` and `set_slice`.

### DIFF
--- a/src/objects/list.rs
+++ b/src/objects/list.rs
@@ -23,7 +23,7 @@ use ffi::{self, Py_ssize_t};
 use conversion::{ToPyObject, ExtractPyObject};
 
 /// Represents a Python `list`.
-pub struct PyList<'p>(PyObject<'p>);
+pub struct PyList<'p>(pub PyObject<'p>);
 
 pyobject_newtype!(PyList, PyList_Check, PyList_Type);
 
--- a/src/objects/list.rs
+++ b/src/objects/list.rs
@@ -17,13 +17,13 @@
 // DEALINGS IN THE SOFTWARE.
 
 use python::{Python, PythonObject, ToPythonPointer};
-use err::{self, PyResult};
+use err::{self, PyErr, PyResult};
 use super::object::PyObject;
 use ffi::{self, Py_ssize_t};
 use conversion::{ToPyObject, ExtractPyObject};
 
 /// Represents a Python `list`.
-pub struct PyList<'p>(pub PyObject<'p>);
+pub struct PyList<'p>(PyObject<'p>);
 
 pyobject_newtype!(PyList, PyList_Check, PyList_Type);
 
@@ -39,6 +39,21 @@ impl <'p> PyList<'p> {
             t
         }
     }
+
+    /// Construct a list from an existing object.
+    #[inline]
+    pub fn from_object(obj: PyObject<'p>) -> PyResult<'p, PyList> {
+        let py = obj.python();
+        let ptr = obj.as_ptr();
+        unsafe {
+            if ffi::PyList_Check(ptr) != 0{
+                Ok(PyList(obj))
+            } else {
+                Err(PyErr::fetch(py))
+            }
+        }
+    }
+
 
     /// Gets the length of the list.
     #[inline]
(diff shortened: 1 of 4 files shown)


dgrunwald:
`get_item`: `PySequence_GetItem` returns a new reference like all the other sequence methods, so you need to use `from_owned_ptr` instead of `from_borrowed_ptr`. Also, should return a `PyResult` as the `PySequence_GetItem` can fail.

Instead of both `length()` and `size()`, why not a `len()` method as is usual for Rust?
Also, `del_slice` is using `i64` while all other functions are using `usize`. I'm not quite sure which type would be the best here -- AFAIK negative arguments are valid for these functions (always with the usual python semantics, so we don't necessarily have to expose this to Rust code since the user can just subtract from `len()` manually).

Safety-wise, I don't think functions like `PySequence_Concat` are guaranteed to return another sequence. We should use `PyObject` as return type, or at least call `PySequence_Check` and return a `TypeError` if a non-sequence is returned.

ehiggs:
Thanks for the feedback.

>get_item: PySequence_GetItem returns a new reference like all the other sequence methods, so you need to use from_owned_ptr instead of from_borrowed_ptr. Also, should return a PyResult as the PySequence_GetItem can fail.

I used the approach used in `PyList`, which is of course wrong since `PyList_GetItem` returns a borrowed reference. But it can also fail and returns `NULL` in the C api; so I asserted here as well. I can try to change it, but there are some lifetime issues that need sorting out in the Iterator impl then.

> Instead of both length() and size(), why not a len() method as is usual for Rust?

Just pedantry to stick to the Python names. I'm happy to change to `len`.

> AFAIK negative arguments are valid for these functions (always with the usual python semantics

Great point. 

>AFAIK negative arguments are valid for these functions (always with the usual python semantics, so we don't necessarily have to expose this to Rust code since the user can just subtract from len() manually).

I'll add more tests.

>Safety-wise, I don't think functions like PySequence_Concat are guaranteed to return another sequence. We should use PyObject as return type, or at least call PySequence_Check and return a TypeError if a non-sequence is returned.

Aside from error handling, I thought concatenation was [closed](https://en.wikipedia.org/wiki/Closure_%28mathematics%29) under sequences. Just as repetition was closed. If you know of some cases where it isn't, I'm happy to change this too. Or I can add the check.

dgrunwald:
>  I used the approach used in PyList, which is of course wrong since PyList_GetItem returns a borrowed reference. But it can also fail and returns NULL in the C api; so I asserted here as well. I can try to change it, but there are some lifetime issues that need sorting out in the Iterator impl then.

`PyList_GetItem` only fails if the index is invalid, but `PySequence_GetItem` can call user-defined `__getitem__` methods and thus fail in arbitrary other ways. So you'll have to return a `PyResult<'p, PyObject<'p>>`.
Now that you mention lifetimes: I don't think lifetime elision is doing the right thing here.  You have to change `PyResult<'p, PySequence>` to `PyResult<'p, PySequence<'p>>`, otherwise the result sequence can only be used while the input sequence is live.

> Just pedantry to stick to the Python names. I'm happy to change to len.

The Python name is also `len()` (or `__len__()`). The CPython name isn't as important as the Python name :)

> Aside from error handling, I thought concatenation was closed under sequences. Just as repetition was closed. If you know of some cases where it isn't, I'm happy to change this too. Or I can add the check.

The sequence protocol can be implemented by user-defined Python classes. `PySequence_Concat` ends up calling the user-defined `__add__` method, and that might return anything. Arguably the user-defined class is broken if it doesn't return another sequence, but since Rust and Python are both (mostly) memory-safe languages, we must avoid undefined behavior even for broken user-defined sequence types.

ehiggs:
I've updated the code and done some cleanup. 

* Changed `length` to `len`.

* Concat and Repeat now return `PyObject`.

* Sizes passed in are signed so people can access from the back of the Sequences.

* `get_item` returns a `PyResult` with a specified lifetime since lifetime elision was not adequate.

* I now use `err::error_if_minusone` where appropriate.

PR #31, opened 2015-09-08
Add Mapping support

ehiggs:
This implements the mapping protocol. As part of testing this, I also added tests for PyDict.

### DIFF
--- a/src/objects/dict.rs
+++ b/src/objects/dict.rs
@@ -212,8 +212,8 @@ mod test {
         let mut v = HashMap::new();
         v.insert(7, 32);
         let dict = v.to_py_object(py);
-        dict.set_item(7i32, 42i32); // change
-        dict.set_item(8i32, 123i32); // insert
+        assert!(dict.set_item(7i32, 42i32).is_ok()); // change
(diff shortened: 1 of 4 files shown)


ehiggs:
This also has support for mapping a BTreeMap to a Dict.

dgrunwald:
* The `_string` methods are redundant: the generic methods already accept `&str` arguments.
* Instead of `result_from_owned_ptr`+`unchecked_cast_into`, it would be safer to use `err::result_cast_from_owned_ptr`.
* `PyMapping_Keys`/`PyMapping_Values`/`PyMapping_Items` are documented to return a list or tuple; you can't assume that it's always a list.

Overall, I don't think it is necessary to provide `PyMapping` in rust-cpython.
* `PyMapping_Check` is unreliable on Python 3
* `len`,`get_item`,`set_item` and `del_item` are already provided by `ObjectProtocol`
* `has_key(k)` is equivalent to `get_item(k).ok()`
* `keys()` is basically equivalent to `call_method("keys", NoArgs, None)` (same with `values()` and `items()`)

I'll merge the `PyDict` tests and `BTreeMap` impl.

PR #41, opened 2016-03-08
Get a mutable reference to Rust objects in Python types

Luthaf:
**[Not ready to merge]**

I am not sure of what I am doing with PyRustObject, but the documentation says:

> Because all Python objects potentially have multiple owners, the concept of Rust mutability does not apply to Python objects. As a result, this API will allow mutating Python objects even if they are not stored in a mutable Rust variable.

Is there any way to create a function with name `$data_name + _mut` in  the `py_clas_impl!` macro?

### DIFF
--- a/src/rustobject/class.rs
+++ b/src/rustobject/class.rs
@@ -30,7 +30,7 @@ pub trait PythonObjectFromPyClassMacro : python::PythonObjectWithTypeObject {
 # Example
 ```
 #[macro_use] extern crate cpython;
-use cpython::{Python, PyResult, PyType, PyDict};
+use cpython::{Python, PyResult, PyType, PyDict, PyObject};
 
 py_class!(class MyType, data: i32, |py| {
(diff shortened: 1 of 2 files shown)


Luthaf:
Another way to do it would be to have `pub fn $data_name` to always return a mutable reference.

dgrunwald:
You cannot hand out a `&mut T` like that. `&mut T` is not only about mutability, but also about uniqueness: there cannot be two `&mut T` instances pointing to the same value.

Your code allows trivially violating that rule:
```
let x = self.inner_mut(py);
let y = self.inner_mut(py);
// we have two aliased &mut references --> undefined behavior
```
If you need mutability, you'll have to use `Cell` or `RefCell`.

Also, `PyRustObject` might be going away soon; I'm currently completely rewriting the "extension class" stuff.

Luthaf:
Ok, thank you for the pointer to `RefCell`, I did not though of that.

PR #46, opened 2016-04-18
Make struct generated with py_class! public

knsd:
So it can be defined not only in `lib.rs`.

### DIFF
--- a/extensions/btree.rs
+++ b/extensions/btree.rs
@@ -14,7 +14,7 @@ py_module_initializer!(btree, initbtree, PyInit_btree, |py, m| {
 
 /// Newtype around PyObject that implements Ord using python value comparisons.
 /// Python exceptions are converted into Rust panics.
-struct OrdPyObject(PyObject);
+pub struct OrdPyObject(PyObject);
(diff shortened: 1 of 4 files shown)


dgrunwald:
I'm not sure if it's a good idea to unconditionally make the type public. I'd like to give the user control over the visibility of the struct. An earlier version of `py_class!` allowed this using the syntax `py_class!{class S}` or `py_class!{pub class S}`, but that feature was lost in the big macro refactoring.
It also leads to some ugly macro code duplication (unless something like [RFC #1575](https://github.com/rust-lang/rfcs/pull/1575) is accepted).

But you don't have to define the type in the crate root (lib.rs): you can also define it in another module, and have that module export a function:
```
pub fn register_type_in_module(py: Python, m: &PyModule) -> PyResult<()> {
    m.add_type::<MyLocalPrivateType>()
}
```

knsd:
Oh, it's great, thanks, Daniel!

I will try to add `py_class!{class S}` or `py_class!{pub class S}` but in another PR, of course.

PR #48, opened 2016-06-02
Add py_exception! macro for defining custom exception types

sciyoshi:
First of all, thank you for this awesome library! This is great for writing native extension modules. This PR adds support for defining custom exception types using `PyErr_NewException` by defining a new macro `py_exception!(modulename, ExceptionName[, BaseExceptionType])`. cc @dgrunwald

### DIFF
--- a/src/err.rs
+++ b/src/err.rs
@@ -24,8 +24,101 @@ use objects::{PyObject, PyType, exc};
 use objects::oldstyle::PyClass;
 use ffi;
 use libc;
+use std::ptr;
+use libc::c_char;
 use conversion::ToPyObject;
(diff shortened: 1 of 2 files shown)


dgrunwald:
Thanks, just two small remarks:
* The `err` and `object` modules aren't supposed to be public (their types are reexported by the main module). Your macro doesn't seem to require these to be public, so can you remove the `pub` modifier again?
* Can you document `PyErr::add_type`?

sciyoshi:
@dgrunwald updated according to your comments - the exported modules were from a WIP version of the code that needed them.

PR #49, opened 2016-06-06
Add support for overloading comparison operators with __richcompare__

sciyoshi:
@dgrunwald this is more generic than `__cmp__`, so I'm not sure if you still want that slot to be implemented as well.


dgrunwald:
On `__cmp__`: I don't think we need that slot. It no longer exists in Python 3.

Cython calls this slot `__richcmp__` instead of `__richcompare__`, it would be nice if we were consistent with that.

On the `CompareOp` enum, how is a user supposed to handle the `Other` case? It might make more sense to leave out this enum variant, and return an error when `op` is invalid without calling `$f`.

Also, can you add some test cases to tests/test_class.rs ?
Especially try comparisons between different types; I think they act similarly to binary operators. That is: Python may call the slot with only one of the operands being of your class type; so you can't rely on `slf` being of type `$class`.

`__richcmp__` should also return `NotImplemented` instead of raising a `TypeError` when the operand types are not supported, so that Python can try calling the other operand's slot. For the binary numeric slots, I've left that task to the user code (and force the operands to be of type `&PyObject`). It would be nice if the macro could do this automatically, but that's not required for a first version of the slot.

sciyoshi:
Thanks for the feedback @dgrunwald

- The `py_class!` method name has been changed to `__richcmp__`.
- I've removed the comment about needing to implement `__cmp__` - it now suggests using `__richcmp__` instead.
- I removed `Other` from the `CompareOp` enum. Because of this, it can no longer implement `From<isize>`, so I've instead added a helper method called `extract_op` in `slots.rs`. Let me know if this makes more sense as an impl on `CompareOp`.
- I've added a few test cases - note that Python always seems to call the comparison function with the object of that type as the first argument, so no special handling is needed here.
- I'm now returning `NotImplemented` in case of a type error. However, user code will still need to return this sentinel value manually - error values won't be converted.

PR #50, opened 2016-06-14
Add support for in-place numeric operators in py_class!

sciyoshi:
@dgrunwald this adds support for all the inplace methods in `PyNumberMethods` (except for `pow()`, which is actually a ternary function).


dgrunwald:
Thanks!

PR #78, opened 2017-01-02
fix #77: make py_class struct public

### LINKED ISSUE #77: pub py_class!
I found this: https://github.com/dgrunwald/rust-cpython/blob/623056d3cafed62fbdfb941ac1fb69225f0b6bce/src/py_class/py_class_impl3.rs#L43

So there is no way to write `py_class` in some file and then import it in another? Because now I have:

```
error: struct `MyType` is private
  --> src/lib.rs:28:5
   |
28 | use self::module::MyType;
   |     ^^^^^^^^^^^^^^^^^^^^^^
```

when `py_class!(class MyType |py| {` is written in `module/mod.rs` file.

Or did I miss something and there is a way to create public class?

pashinin:


dgrunwald:
I don't think it's a good idea to make the type generated by `py_class` public unconditionally. I've now added the `py_class!(pub class ...)` syntax so that the visibility can be selected when calling the macro.

Issue #39, opened 2016-03-07
Flexible parameter lists


dgrunwald:
Add support for:

- [ ] Optional parameters
- [ ] `*args`
- [ ] `**kwargs`

Luthaf:
Did you had the time to implement this? The [documentation](https://dgrunwald.github.io/rust-cpython/doc/cpython/macro.py_argparse!.html) say so, and parameters with or without explicit type works.

But trying to use default values gives me errors about unexpected value:

```rust
#[macro_use] extern crate cpython;
use cpython::{Python, PyResult};

py_class!(class MyType |py| {
    data number: i32;
    def __new__(_cls, arg: i32) -> PyResult<MyType> {
        MyType::create_instance(py, arg)
    }
    def foo(&self, a, b: i32 = 8) -> PyResult<PyObject> {
    // ===> Error: no rules expected the token `8`at line 62 col 44 in <cpython macros>
        println!("{} {}", a, b);
        Ok(py.None())
    }
});
```

I tried to read the code to see where this could come from, but my macro-fu is not that good ...

PR #81, opened 2017-01-23
Implement optional values in argparse

Luthaf:
I was missing this from #39, and I finally got my head around the code, so here is an implementation of the optional parameters in parameters lists.

I am not sure I understood all the interaction between all the `*Object` traits, so the code might not be optimal. In particular, I don't really get what this line is doing, and if it is the best way to do it:

```rust
$iter.next().unwrap().as_ref().unwrap_or(<$ptype as $crate::ToPyObject>::into_py_object($default, $py).as_object())
```

Tell me what you think!

### DIFF
--- a/src/argparse.rs
+++ b/src/argparse.rs
@@ -258,8 +258,17 @@ macro_rules! py_argparse_parse_plist_impl {
             ($($tail)*)
         }
     };
+    // Optional parameter with reference extraction
+    { $callback:ident $initial_args:tt [ $($output:tt)* ]
+        ( $name:ident : &$t:ty = $default:expr, $($tail:tt)* )
+    } => {
+        py_argparse_parse_plist_impl! {
+            $callback $initial_args
+            [ $($output)* { $name:&$t = [ {} {$default} {$t} ] } ]
+            ($($tail)*)
+        }
+    };
     // Optional parameter
-    // TODO: with reference extraction?
     { $callback:ident $initial_args:tt [ $($output:tt)* ]
         ( $name:ident : $t:ty = $default:expr , $($tail:tt)* )
     } => {
@@ -353,8 +362,8 @@ macro_rules! py_argparse_param_description {
             is_optional: false
         }
     );
-    // optional parameter with a default value
-    { $pname:ident : $ptype:ty = [ {} {$default:expr} {} ] } => (
+    // optional parameters
+    { $pname:ident : $ptype:ty = [ {} {$default:expr} {$($rtype:tt)*} ] } => (
         $crate::argparse::ParamDescription {
             name: stringify!($pname),
             is_optional: true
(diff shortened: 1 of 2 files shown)


dgrunwald:
[parse_args()](https://dgrunwald.github.io/rust-cpython/doc/cpython/argparse/fn.parse_args.html) matches the arguments with the parameters, and writes its result into an array of type `[Option<PyObject>]`, where each array element corresponding to one parameter in the function signature. If the option is `None`, that indicates that the parameter is optional and was not passed an argument.

Ideally, we'd then assign each array element to a variable named like the parameter, e.g.:
`let [ref p1, ref p2, ref p3] = output;`
(which could be done in a macro as: `let [$(ref $pname),*] = output;`)
Unfortunately, slice patterns are not available in stable rust, so we have to use another approach to get at the array elements. My solution to this problem was to use an iterator instead.

As `argparse!()` can be used within any function of any return type, we can't use `try!()` for error handling, so the macro needs to generate nested matches to handle the extraction errors. This is why the recursive `py_argparse_extract!` helper macro exists.

So `iter.next().unwrap()` is a reference to an element of the `output` array, of type `&Option<PyObject>`. The `unwrap` is guaranteed to succeed because we know the length of `output` array matches the number of recursions of the `py_argparse_extract!` helper macro.

`.as_ref()` then turns this into `Option<&PyObject>`.
Your `unwrap_or()` call then constructs a new Python object to use as a default value if the Option is `None`.

Finally, the `extract` call converts the Python object into the expected type for the Rust parameter.
So if the default value is used, you have an unnecessary `Rust -> Python -> Rust` conversion. It would be better to call `extract()` only if the Option was `Some`, and directly use the Rust default value otherwise.

Maybe something like: `iter.next().unwrap().as_ref().map(|obj| extract<...>(py, obj)).unwrap_or(Some($default)).`

dgrunwald:
Also, there's two different code paths for extracting values and extracting references. Depending on whether the first token in the type name is `&`, the macro `py_argparse_parse_plist_impl!` will set the `$rtype` element in the `detail` tt (to the reference's target type), or leave it empty (if there was no `&`).

Full form of a parsed parameter tt:
```
{ $pname:ident : $ptype:ty = $detail:tt }
{ $pname:ident : $ptype:ty = { $kwargs:tt $default:tt $rtype:tt } }
{ $pname:ident : $ptype:ty = { {**} {$default:expr} {$rtype:ty} } }
```

We might want to support default values in the codepath for reference extraction (non-empty $rtype) too. This would allow using a signature like: `(name: &str = "")`

Luthaf:
> Maybe something like: iter.next().unwrap().as_ref().map(|obj| extract<...>(py, obj)).unwrap_or(Some($default)).

This is done, thank you for the explanations.

> We might want to support default values in the codepath for reference extraction (non-empty $rtype) too. This would allow using a signature like: (name: &str = "")

I have implemented this, but the code does to Rust => Python => Rust round trip for now. I tried to use the same trick as for the non-reference types, but the compiler is complaining about a missing implementation of `FromPyObject`:

```
match $iter.next().unwrap().as_ref().map(|obj| obj.extract::<_>($py)).unwrap_or(Ok($default)) {
                                                   ^^^^^^^ the trait `conversion::FromPyObject<'_>` is not implemented for `&str

```

PR #98, opened 2017-03-25
add PyAsyncMethods support

fafhrd91:


fafhrd91:
code seems to work 

here is example
https://github.com/fafhrd91/async-tokio/blob/master/tests/test_basic.py#L132
https://github.com/fafhrd91/async-tokio/blob/master/ext/src/future.rs#L325

fafhrd91:
ping?

AraHaan:
Seems that somehow travis is failing...

PR #101, opened 2017-04-07
Python class properties

fafhrd91:
Added class properties via PyGetSetDef

property definition looks like this:

```rust


py_class!(class ClassWithProperties |py| {
    data num: Cell<i32>;

    property DATA {
        get(&slf) -> PyResult<i32> {
            Ok(slf.num(py).get())
        }
        set(&slf, value: i32) -> PyResult<()> {
            slf.num(py).set(value);
            Ok(())
        }
    }
});
```

in this implementation it is not possible to use `&self`

### DIFF
--- a/src/py_class/mod.rs
+++ b/src/py_class/mod.rs
@@ -23,6 +23,7 @@ mod py_class_impl2;
 mod py_class_impl3;
 #[doc(hidden)] pub mod slots;
 #[doc(hidden)] pub mod members;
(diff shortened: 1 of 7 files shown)


fafhrd91:
Also, getter and setter is not available in rust. This and availability are related how rust macros works. Another option would be allow user define unique names for getter and setter

fafhrd91:
@dgrunwald ping

PR #103, opened 2017-04-28
add PySlice support

fafhrd91:


fafhrd91:
@dgrunwald are you planing to work on rust-cpython?

PR #105, opened 2017-04-28
test that PyBuffer::copy_from_slice copies data from buffer

astraw:
Hi, as far as I can tell, this is the correct test. (Or do I understand the code wrong?)


astraw:
Oh, sorry my mistake I didn't notice that this is a test that the buffer should be read-only.

PR #111, opened 2017-09-30
Update the regex dependency

Eijebong:
The `.at` method has been replaced by `.get` and now returns a `Match`.


Eijebong:
Build failure seems to be unrelated

PR #116, opened 2017-11-16
Update builds for latest nightly

markbt:
This is a collection of fixes that should make the nightly CI pass.


markbt:
The AppVeyor build failed because it failed to download nightly.  I think it needs to be kicked off again and then should work.

Issue #125, opened 2018-02-21
Implement Error trait for PyErr


asoderman:
This would allow for easy conversion between error types within a rust program. See: [this](https://doc.rust-lang.org/book/first-edition/error-handling.html#the-error-trait)

PR #126, opened 2018-02-21
Implemented Error trait for PyErr

asoderman:
This allows for more flexibility for converting between error types. 

Relevant issue: #125


src/err.rs
```diff
@@ -394,6 +396,21 @@ impl <'p> std::convert::From<PythonObjectDowncastError<'p>> for PyErr {
     }
 }
 
+impl Error for PyErr {
+
```
Eijebong: :scissors:

src/err.rs
```diff
@@ -394,6 +396,21 @@ impl <'p> std::convert::From<PythonObjectDowncastError<'p>> for PyErr {
...
+    fn description(&self) -> &str {
+        "Error originating from the rust-cpython bindings."
+    }
+
+}
+
+impl fmt::Display for PyErr {
+
```
Eijebong: :scissors:

asoderman:
Fixed

Ploppz:
Waiting for this to be merged. Is it because Travis failed? Seems like it cannot compile some test - but this seems to not be caused by this PR (?)

asoderman:
I sync'd my branch with master and it appears to have fixed the error with the doctest however there is a new error with nightly. Not really sure what is causing it.

Ploppz:
Still waiting for this one... @asoderman can you perhaps rebase and let's see if CI issues have been resolved?

src/err.rs
```diff
@@ -394,6 +396,18 @@ impl <'p> std::convert::From<PythonObjectDowncastError<'p>> for PyErr {
...
 
+impl Error for PyErr {
+    fn description(&self) -> &str {
+        "Error originating from the rust-cpython bindings."
+    }
+}
+
+impl fmt::Display for PyErr {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "PyErr: ptype {:?} pvalue {:?}", self.ptype, self.pvalue)
```
dgrunwald, line 407: I would expect Display to result in something like `str(e)` in Python.
This code here is more like `repr(type(e))` + `repr(e)`, except even weirder if the exception is not yet initialized (`PyErr::new_lazy_init`).

Issue #102, opened 2017-04-12
Document py27 support in README


jbremer:
Probably would be useful to mention that Python 2.7 is disabled by default in the README and/or documentation. I found the relevant options in one of the tests, which is fine, but some more verbosity could be useful rather than `_Py_deinitialize not found` (or whatever the symbol was) when trying to `import library` from the Python interpreter.

markbt:
Fixed in #128

Issue #74, opened 2016-12-30
Document Cargo.toml features


ssokolow:
When looking at the README, I almost concluded that rust-cpython would auto-detect the installed major Python version and didn't allow specifying which to use when 2.x and 3.x were both found.

Furthermore, once I *did* figure out that it uses features to select them, I still couldn't find them mentioned anywhere in the documentation. That's not exactly ergonomic.

...and, as someone who uses Python heavily but has almost no experience with C or C++, I think the description of the `extension-module` feature would benefit from going into a bit more detail.

(eg. I *think* an "extension module" is a "compiled module to be loaded into a Python program using `import`), but I'm not feeling very confident beyond a vague memory of seeing the term when I binge-read the entire distutils section of the Python manual and then poked around in the distutils code in order to figure out how to accomplish certain tasks... definitely not something many potential users will have done.)

ssokolow:
@forensicmatt 
If I understand the `Cargo.toml` for the project correctly, tell cargo to build/run your project with the `python27-sys` feature and that'll cause this to depend on Python 2.7.

forensicmatt:
I figured it out by reading up on the choosing features section in the [Specifying Dependencies docs](http://doc.crates.io/specifying-dependencies.html#choosing-features). Here is how I specified just python27 in case it helps someone else.
```
[package]
name = "pytest"
version = "0.1.0"
authors = ["test author"]

[lib]
name = "pytest"
crate-type = ["cdylib"]

[dependencies.cpython]
version = "*"
default-features = false
features = ["python27-sys"]
```

AaronM04:
Is there any particular reason this example isn't also in README.md? If not, I can make a PR.

markbt:
The docs now list how to enable python2.7 support.

PR #128, opened 2018-03-04
Added docs for python 2.7 feature 

rminderhoud:
See issues #102 & #74


markbt:
Thanks for the contribution.  This would be good to get documented.  You need to mark the docblock as `ignore` to stop rust from trying to run it as a test.  See [this commit](https://github.com/dgrunwald/rust-cpython/commit/51dea15cac60f9bcf81f7554eefd1afd29dee2eb) where I fix it for another case.  That should make the CI pass.

rminderhoud:
Ah sorry, I've made the change and rebased

PR #131, opened 2018-03-28
Feature/python as crate

svevang:
This PR demos a new scheme to package python code up as a crate.  The end goal is to simplify development in Rust. Developers can mix and match dependencies that call into python, and the tooling this PR adds to rust-cpython would (theoretically) take care of the rest.  Please take note that this PR feature is a prototype, but I wanted to open it anyways in order to gauge interest and get some feedback that would be incorporated into any final feature.

What's added here is a way to build and link python using Cargo. The excellent `python-build` tool in the [pyenv plugins](https://github.com/pyenv/pyenv/tree/master/plugins/python-build) is invoked to build a copy of python in the new `python-sys` crate (added to the cpython workspace). The resulting python interpreter and dynamic library are handed off to the python3-sys crate and the building and linking proceeds as normal. Python 2.7 support is left as a TODO

To see this in action run:
```
±  |feature/python-as-crate ✓| → cargo build --features build-python-3-5-0

```
### Example crate
I created an example over here:
https://github.com/svevang/pipclient-tokenize

This is a simple rust crate that packages up python code--allowing rust to call into NLTK functions. Briefly, here's how we set it up:
First select a python version using a new `build-python-*` feature:

```
# inside Cargo.toml
[dependencies.cpython]
   git = "https://github.com/svevang/rust-cpython.git"
  features = ["build-python-3-5-0"]
# needed because python-sys is not its own crate
[build-dependencies.cpython]
  git = "https://github.com/svevang/rust-cpython.git"
  features = ["build-python-3-5-0"]
 ```

Then inside your `build.rs` file, install your python code using pip:

```
cpython::pip_helpers::install_python_package(python_code_path, Some(requirements_path));
```

Then inside your lib.rs you can interact with your python module in the usual rust-cpython way.  That's it!

As mentioned earlier, this adds a new `python-sys` crate, which could be a good way to bootstrap the eventual merger of `python3-sys` and `python27-sys`. 

#### MacOS
The build seems pretty stable for me on MacOS, but if you encounter build errors please check out https://github.com/pyenv/pyenv/wiki/Common-build-problems

#### Linux
Because Linux doesn't set up rpaths in the resulting binaries, you need LD_LIBRARY_PATH for the final build command invocation, which kinda defeats the purpose of this feature. The example app builds but you need to supply the newly compiled python's dynamically dylib location:

```
root@4e7939d4d074:~/pipclient-tokenize# ./target/debug/main
./target/debug/main: error while loading shared libraries: libpython3.5m.so.1.0: cannot open shared object file: No such file or directory

root@4e7939d4d074:~/pipclient-tokenize# LD_LIBRARY_PATH=/root/pipclient-tokenize/target/debug/build/python-sys-7ca8397d2e539fb3/out/python3.5.0-build/lib/ ./target/debug/main 
Result: Ok(["Hello", ",", "how", "are", "you", "?", "I", "am", "fine", "thank", "you", "."])
```

Here's the Dockerfile that I'm using to test Linux with:
https://gist.github.com/svevang/e97e63c8147c2f4c01711a79d6a4f7d9

Any suggestions or workarounds for getting python to link without setting env vars would be appreciated!

#### Windows
I've not tested this on Windows, but it seems very unlikely to work. I gather you can install a posix subsystem in windows now (including bash), so maybe the python-build scripts would work? Likely there would be problems with that and with linkers as well.

### CODE REVIEW

File: python3-sys/build.rs [hunk 1]
```diff
@@ -388,6 +389,10 @@ fn main() {
     // If you have troubles with your shell accepting '.' in a var name, 
     // try using 'env' (sorry but this isn't our fault - it just has to 
     // match the pkg-config package name, which is going to have a . in it).
+    match python_sys::python_path() {
+        Some(path) => env::set_var("PYTHON_SYS_EXECUTABLE", path),
```
svevang: This is a ugly way to hook into the build-time configuration of the python3-sys library. The preferable way to do this would be to refactor how we use [`fine_interpreter_and_get_config`](https://github.com/dgrunwald/rust-cpython/pull/131/files#diff-646601a8a4d8d5c828181358c7c201ceL254)-- only calling that method if we searching for a python install. Presumably if we're building and supplying our own python, then we should also be able to take our build time configuration and patch through those variables supplied by this build script (e.g. `Py_USING_UNICODE`).

svevang:
Closing this as it’s not really mergable. Todo are the Linux linking problems and breaking this PR up into more bite size chunks.

PR #150, opened 2018-08-29
fix #144 python 3.7 change the way to work with threads

### LINKED ISSUE #144: Python 3.7.0 support
Since python 3.7 threading is not optional,
for example from [here](https://docs.python.org/3/library/threading.html)

> Changed in version 3.7: This module used to be optional, it is now always available.

So it tells that no `WITH_THREAD`, but actually threading is enabled:
```sh
Python 3.7.0 (default, Jul 15 2018, 10:44:58) 
[GCC 8.1.1 20180531] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sysconfig
>>> config = sysconfig.get_config_vars()
>>> print(config['WITH_THREAD'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'WITH_THREAD'
>>> 
```

and compilation of 1bdca2236fa832d3c893c40d7d180cccd9ae8ea2 failed on arch linux/x86_64
```
$ cargo build
   Compiling python3-sys v0.2.0 (file:///home/buildbot/rust-cpython/python3-sys)
   Compiling cpython v0.2.0 (file:///home/buildbot/rust-cpython)
warning: unused import: `c_uchar`
 --> python3-sys/src/code.rs:1:20
  |
1 | use libc::{c_void, c_uchar, c_char, c_int};
  |                    ^^^^^^^
  |
  = note: #[warn(unused_imports)] on by default

error[E0425]: cannot find function `PyEval_ThreadsInitialized` in module `ffi`
  --> src/pythonrun.rs:55:26
   |
55 |             assert!(ffi::PyEval_ThreadsInitialized() != 0);
   |                          ^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `ffi`

```

to fix build I have to disable `WITH_THREAD` check:
```diff
diff --git a/python3-sys/src/ceval.rs b/python3-sys/src/ceval.rs
index ab67b12..b85f6a2 100644
--- a/python3-sys/src/ceval.rs
+++ b/python3-sys/src/ceval.rs
@@ -51,7 +51,6 @@ pub unsafe fn PyEval_CallObject(callable: *mut PyObject, arg: *mut PyObject) ->
     pub fn PyEval_RestoreThread(arg1: *mut PyThreadState) -> ();
 }
 
-#[cfg(py_sys_config = "WITH_THREAD")]
 #[cfg_attr(windows, link(name="pythonXY"))] extern "C" {
     pub fn PyEval_ThreadsInitialized() -> c_int;
     pub fn PyEval_InitThreads() -> ();
```

Dushistov:


Dushistov:
> Did Python 3.7 really change anything here? The PyEval_ThreadsInitialized function still seems to exist 

It exists, but assert in `cpython/src/pythonrun.rs`:
```Rust
// If Python isn't initialized yet, we expect that Python threading isn't initialized either.
assert!(ffi::PyEval_ThreadsInitialized() == 0);
```
failed, because of with python 3.7 `ffi::Py_IsInitialized` initializes `GIL`,
so almost all unit tests failed. Because of I don't see the way to check python version in `cpython` to disable assert for `3.7`, I implement specific version of `PyEval_ThreadsInitialized ` for 3.7,
the other way (if `cfg(Py_3_7)` will be available in `cpython`) may be:

```Rust
// If Python isn't initialized yet, we expect that Python threading isn't initialized either.
    if cfg!(not(Py_3_7)) {
            assert!(ffi::PyEval_ThreadsInitialized() == 0);
    }
```

dgrunwald:
Digging into the Python source code, it looks like in 3.7, `PyEval_ThreadsInitialized` initially returns true (when Python is completely uninitialized), then starts returning false while Python is initializing, then true again when the Python (including the GIL) is fully initialized.

I think we'll just need to avoid calling anything other than `Py_IsInitialized()` prior to the `Py_InitializeEx()` call when targeting Python 3.

In Python 2 the situation is different, as back then it was possible to initialize the GIL without initializing the Python interpreter. (the documentation explicitly allowed calling `PyEval_InitThreads` prior to `Py_InitializeEx`)

dgrunwald:
I've fixed this in master by making the assertion Python2-only.

Issue #141, opened 2018-07-19
Compiling front-page example fails on OSX


PureW:
I'm trying to compile the example in the root-level README and it's failing with linker errors.

I summed it up in this example code here, re-run it to repeat issues:

```
#!/usr/bin/env bash

set -ex

python3 -m venv venv
source venv/bin/activate

set -eux

path="example"

rm -rf $path venv


cargo new --lib $path

cat >> $path/Cargo.toml << EOL
[lib]
name = "rust2py"
crate-type = ["cdylib"]

[dependencies.cpython]
version = "0.1"
features = ["extension-module"]
EOL

cat > $path/src/lib.rs << EOL
#[macro_use] extern crate cpython;

use cpython::{PyResult, Python};

py_module_initializer!(librust2py, initlibrust2py, PyInit_librust2py, |py, m| {
    try!(m.add(py, "__doc__", "This module is implemented in Rust."));
    try!(m.add(py, "sum_as_string", py_fn!(py, sum_as_string_py(a: i64, b:i64))));
    Ok(())
});

// logic implemented as a normal rust function
fn sum_as_string(a:i64, b:i64) -> String {
    format!("{}", a + b).to_string()
}

fn sum_as_string_py(_: Python, a:i64, b:i64) -> PyResult<String> {
    let out = sum_as_string(a, b);
    Ok(out)
}
EOL

cd $path
cargo build
```

Part of errors:
```
...
  = note: Undefined symbols for architecture x86_64:
            "_PyExc_UnicodeDecodeError", referenced from:
                _$LT$cpython..objects..exc..UnicodeDecodeError$u20$as$u20$cpython..python..PythonObjectWithCheckedDowncast$GT$::dow
ncast_from::h41d3f459c795f26d in libcpython-51dd2553a386907e.rlib(cpython-51dd2553a386907e.cpython1.rcgu.o)
            "_PyExc_TypeError", referenced from:
                _$LT$cpython..objects..exc..TypeError$u20$as$u20$cpython..python..PythonObjectWithTypeObject$GT$::type_object::h32e
17c3d5136b841 in libcpython-51dd2553a386907e.rlib(cpython-51dd2553a386907e.cpython1.rcgu.o)
            "_PyExc_OverflowError", referenced from:
                _$LT$cpython..objects..exc..OverflowError$u20$as$u20$cpython..python..PythonObjectWithTypeObject$GT$::type_object::
[...]

PR #154, opened 2018-10-05
Update README with OsX build-instructions

PureW:
Fixes https://github.com/dgrunwald/rust-cpython/issues/141


dgrunwald:
Is there a way to automatically pass these linker arguments when the `extension-module` feature is in use? For example from `build.rs`?

Issue #159, opened 2018-10-26
use new (1.30.0) stable rust feature `#[macro_export(local_inner_macros)]`


softprops:
announced [here](https://blog.rust-lang.org/2018/10/25/Rust-1.30.0.html) rust is updating the way [users access exported macros](https://blog.rust-lang.org/2018/10/25/Rust-1.30.0.html#use-and-macros) and the way macro authors export them noted in the [release notes](https://github.com/rust-lang/rust/blob/master/RELEASES.md#language)

> You can now import and reexport macros from other crates with the use syntax. Macros exported with #[macro_export] are now placed into the root module of the crate. If your macro relies on calling other local macros, it is recommended to export with the #[macro_export(local_inner_macros)] attribute so users won't have to import those macros.

pending docs for this feature live [here](https://github.com/rust-lang-nursery/edition-guide/pull/117/files#diff-4e57182b623c416bc83329f6a7307ec9R84)

I've started experimenting with this in the rust crowbar crate but have found that in practice it may require this crate to use the `local_inner_macros` way of exporting macros so crowbar users won't have to `use` this crates helper macros explicitly https://github.com/ilianaw/rust-crowbar/pull/46#issuecomment-433458754

softprops:
since there are so many of these it may be just as simple to use the new `$crate::`  path references within macro definitions.

softprops:
@dgrunwald I don't mean to be a bother, but I noticed there are some passing prs with no activity in this repo since 2017. Is this repo still actively maintained?

dgrunwald:
@softprops No, it's not really maintained anymore.

Probably most of the macro stuff should be re-done using procedural macros instead; especially `py_class!` is way too complex currently and the error messages from `macro_rules` are useless.

softprops:
@dgrunwald that's for the response. Context for my interest here is I'm trying to work with the maintainer of the rust [crowbar crate](https://github.com/ilianaw/rust-crowbar) move forward with the newly stable macro features but it depended on this crate so I reached out. I'll see what we can do!

derekdreery:
I have a PR that makes it work with 2018 edition: https://github.com/dgrunwald/rust-cpython/pull/167.

PR #160, opened 2018-10-26
remove usages of try! macro

softprops:
while starting to look at https://github.com/dgrunwald/rust-cpython/issues/159 I noticed an enormous amount of try! macro usages.  The docs for [try](https://doc.rust-lang.org/std/macro.try.html) recommend using `?` instead. I thought I might take a look at this before taking a stab at a bigger pull for the 1.30 macro updates


softprops:
`?` was introduced way back in [1.13](https://blog.rust-lang.org/2016/11/10/Rust-1.13.html)

softprops:
@dgrunwald friendly ping

PR #161, opened 2018-11-05
build.rs: Be resilient to the absence of "python" executable

tsnoam:
Fix extreme case where attempting to build on a machine which only has
"pythonX" or "pythonX.Y" executable.

Previously, if the "python" executable didn't exist on the system,
attempting to call get_config_from_interpreter("python")
resulted in the entire find_interpreter_and_get_config() function to
return Error.
With this change, the abscense of "python" executable will be ignored
and the other executable names will be attempted before giving up.

### DIFF
--- a/python27-sys/build.rs
+++ b/python27-sys/build.rs
@@ -277,28 +277,19 @@ fn find_interpreter_and_get_config(expected_version: &PythonVersion) ->
                                interpreter_version));
         }
     }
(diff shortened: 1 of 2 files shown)


tsnoam:
Hi @dgrunwald 
Just a kind ping. Have you had any chance to look at this PR?

Thanks!

PR #162, opened 2018-11-24
Add missing bindings

indygreg:
I've been playing around with Rust applications embedding Python 3 interpreters. There were some APIs missing from python3-sys that I found myself needing. The commits in this series introduce various APIs.

I'm not 100% confident I got some style things correct. But that's what review is for :)

### DIFF
--- a/python3-sys/src/fileutils.rs
+++ b/python3-sys/src/fileutils.rs
@@ -1,0 +1,8 @@
+use libc::{c_char, size_t, wchar_t};
+
+#[cfg(not(Py_LIMITED_API))]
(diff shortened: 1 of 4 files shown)


dgrunwald:
`Py_DecodeLocale` is new in Python 3.5, so you'll need a #[cfg] on the function or module definition.
Otherwise the nightly build complains that python3-sys defines a symbol not found in the .so (if the build wasn't already failing due to another error...)

indygreg:
The Rust Nightly failures appear to be due to problems installing Rust and aren't related to this PR.

The missing `#[cfg]` is legitimate and I'll force push an update shortly.

PR #165, opened 2018-12-04
Basic implementation for Python sets

gracinet:
This exposes the functions that are declared in ffi::setobject.
Since that latter doesn't declare iterators nor update (perhaps for
the same consistency reasons as for dicts ?), we don't have iteration
on Python sets (hence no listing as Vec), nor `update()`.

But this is already enough for Rust extensions to create Python sets
(our primary use-case) and to check membership.

### DIFF
--- a/src/objects/set.rs
+++ b/src/objects/set.rs
@@ -1,0 +1,269 @@
+// Copyright (c) 2018 Daniel Grunwald, Georges Racinet
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
+// software and associated documentation files (the "Software"), to deal in the Software
+// without restriction, including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons
+// to whom the Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+// PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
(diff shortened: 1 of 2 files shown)


gracinet:
Seems like the Travis build failure on rust-nightly is an unrelated rustup problem.

src/objects/set.rs
```diff
@@ -0,0 +1,257 @@
+// Copyright (c) 2018 David Grunwald, Georges Racinet
```
markbt: His name is `Daniel`, not `David`. :)

src/objects/set.rs
```diff
@@ -0,0 +1,257 @@
...
+        pub fn empty(py: Python) -> PyResult<PySet> {
+        unsafe {
+            err::result_cast_from_owned_ptr(py,
+                ffi::PySet_New(ptr::null_mut()))
+        }
+    }
+
+    /// Empty an existing set of all values.
+    #[inline]
+    pub fn clear(&self, py: Python) -> PyResult<()> {  // NOCOMMIT check CPython doc for signature explanation (dict's is different)
```
markbt: Have you got a resolution for this?

gracinet: Yes, I checked the signature difference, it's perfectly correct.
(this is usually catched by my precommt hooks if I don't forget to activate them)

src/objects/set.rs
gracinet: Thanks, seems I've been sloppy when I decided that it wasn't right to put only my own for code so much similar to `dict.rs`.

gracinet:
@markbt thanks for the review ! I added a unit test for good measure for `clear()`.
I can squash and force-push myself if you like.

src/objects/set.rs
```diff
@@ -0,0 +1,269 @@
...
+
+    /// Pop a value
+    /// This is equivalent to the Python expression `self.pop(value)`.
+    /// We get KeyError if the set is empty
+    pub fn pop(&self, py: Python) -> PyResult<PyObject> {
+        let as_opt = unsafe {
+            PyObject::from_borrowed_ptr_opt(py,
+                ffi::PySet_Pop(self.0.as_ptr()))
+        };
+        match as_opt {  // TODO use one of the magics
```
markbt: I don't think you need to use one of the utilities for the error.  This code is clear enough.  Just drop the TODO.

gracinet:
@markbt rebase/squash done, and indeed that makes Travis happy.

PR #166, opened 2018-12-08
Use Rust docstring for __doc__ of instance methods

quark-zju:
This patch allows `__doc__` of instance methods to be specified using Rust
docstrings, like:

    py_class!(... {
        ///Write something.
        ///Second line.
        def foo(&self, ...) -> ... {
            ...
        }
    }

Then `instance.foo.__doc__` would be `"Write something.\nSecond line.\n"` in the
Python world.

Under the hood, Rust translates `/// foo` to `#[doc="foo"]` automatically. So
they can be matched using macros. See https://doc.rust-lang.org/beta/rustdoc/the-doc-attribute.html

### DIFF
--- a/src/py_class/py_class_impl.py
+++ b/src/py_class/py_class_impl.py
@@ -442,7 +442,12 @@ def generate_instance_method(special_name=None, decoration='',
         slot=None, add_member=False, value_macro=None, value_args=None):
     name_pattern = special_name or '$name:ident'
     name_use = special_name or '$name'
-    def impl(with_params):
+    def impl(with_params, with_docs):
+        if with_docs:
+            doc_prefix = '$(#[doc=$doc:expr])*'
+            value_suffix = ', { _cpython__py_class__py_class_impl__concat!($($doc, "\\n"),*) }'
+        else:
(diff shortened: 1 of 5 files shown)


src/py_class/py_class_impl.py
```diff
@@ -442,31 +442,39 @@ def generate_instance_method(special_name=None, decoration='',
...
         if add_member:
             members.append((name_use, value))
         generate_case(pattern, new_impl=impl, new_slots=slots, new_members=members)
-    impl(False) # without parameters
-    impl(True) # with parameters
+
+    # Currently, only instance method implements docs
+    with_docs = (value_macro == 'py_class_instance_method')
+    impl(False, with_docs) # without parameters, without docs
+    impl(True, with_docs) # with parameters, without docs
```
markbt: These comments should probably say "with or without docs"

quark-zju:
Hi, @markbt, Sorry for the late update. But I think this is ready for consideration again.

AdamRzepka:
Hi! Any news on that? Or are there some workarounds available for now, to add docstrings to method?

PR #167, opened 2019-01-02
Make macros work with rust 2018

derekdreery:
I've updated the macros to work with rust 2018, following the [edition guide](https://rust-lang-nursery.github.io/edition-guide/rust-2018/macros/macro-changes.html).


markbt:
These macros add a lot of noise, but they can be removed when we retarget 1.30+ as the base version, so I think they're ok for now.

PR #169, opened 2019-01-10
High level Capsule support

gracinet:
Python Capsules are the preferred way to export / import C-level APIs between extension modules (see https://docs.python.org/3/c-api/capsule.html). I believe them to be key for incremental switch to Rust of existing C extensions.

The first commit introduces the `PyCapsule` struct (based on symbols already provided by `python-sys`), the two subsequent ones add macros that take care of the direct function pointer case and caching using `sync::Once` on a `static mut`

### DIFF
--- a/src/objects/capsule.rs
+++ b/src/objects/capsule.rs
@@ -13,6 +13,153 @@ pub struct PyCapsule(PyObject);
 
 pyobject_newtype!(PyCapsule, PyCapsule_CheckExact, PyCapsule_Type);
 
+
+/// Macro to retrieve a Python capsule pointing to an array of data, with a layer of caching.
+///
+/// For more details on capsules, see [`PyCapsule`]
+///
+/// The caller has to define an appropriate `repr(C)` `struct` first, and put it in
+/// scope (`use`) if needed along the macro invocation.
+///
+/// The arguments of this macro are
+/// - the segments of the full dotted name of the targetted capsule,
+/// - `$rustmod`, a suitable module name,
+/// - `$ruststruct`, the name of the target `struct`.
+///
+/// The macro defines a Rust module with a retrieval function such as:
+///
+/// ```ignore
+/// mod $rustmod {
+///     pub unsafe fn retrieve<'a>(py: Python) -> PyResult<&'a $ruststruct> { ... }
+/// }
+/// ```
+///
+/// The `retrieve()` function is unsafe for the same reasons as [`PyCapsule::import_data`],
+/// upon which it relies.
+///
+/// # Examples
+/// This retrieves and uses one of the simplest capsules in the Python standard library, found in
+/// the `unicodedata` module. The C API enclosed in this capsule is the same for all Python
+/// versions supported by this crate.
+///
+/// In this case, as with all capsules from the Python standard library, the capsule data
+/// is an array (`static struct`) with constants and function pointers.
+/// ```
+/// #[macro_use] extern crate cpython;
+/// extern crate libc;
+///
+/// use cpython::{Python, PyCapsule};
+/// use libc::{c_char, c_int};
+/// use std::ffi::{c_void, CStr, CString};
+/// use std::mem;
+/// use std::ptr::null_mut;
+///
+/// #[allow(non_camel_case_types)]
+/// type Py_UCS4 = u32;
+/// const UNICODE_NAME_MAXLEN: usize = 256;
+///
+/// #[repr(C)]
+/// pub struct unicode_name_CAPI {
+///     // the `ucd` signature arguments are actually optional (can be `NULL`) FFI PyObject
+///     // pointers used to pass alternate (former) versions of Unicode data.
+///     // We won't need to use them with an actual value in these examples, so it's enough to
+///     // specify them as `*mut c_void`, and it spares us a direct reference to the lower
+///     // level Python FFI bindings.
+///     size: c_int,
+///     getname: unsafe extern "C" fn(
+///         ucd: *mut c_void,
+///         code: Py_UCS4,
+///         buffer: *mut c_char,
+///         buflen: c_int,
+///         with_alias_and_seq: c_int,
+///     ) -> c_int,
+///     getcode: unsafe extern "C" fn(
+///         ucd: *mut c_void,
+///         name: *const c_char,
+///         namelen: c_int,
+///         code: *mut Py_UCS4,
+///     ) -> c_int,
+/// }
+///
+/// #[derive(Debug, PartialEq)]
+/// pub enum UnicodeDataError {
+///     InvalidCode,
+///     UnknownName,
+/// }
+///
+/// impl unicode_name_CAPI {
+///     pub fn get_name(&self, code: Py_UCS4) -> Result<CString, UnicodeDataError> {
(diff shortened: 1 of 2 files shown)


src/objects/capsule.rs
```diff
@@ -0,0 +1,548 @@
...
+/// [`PyCapsule`]: struct.PyCapsule.html
+#[macro_export]
+macro_rules! py_capsule_fn {
+    ($($capsmod:ident).+, $capsname:ident, $rustmod:ident, $( $sig: tt)* ) => (
+        mod $rustmod {
+            use super::*;
+            use std::sync::Once;
+            use $crate::PyClone;
+
+            pub type CapsFn = unsafe extern "C" fn $( $sig )*;
```
markbt: No need to abbreviate `Capsule` here.  Calling this `CapsuleFn` would be clearer, I think.

src/objects/capsule.rs
```diff
@@ -0,0 +1,548 @@
...
+}
+
+/// Macro to retrieve a function pointer capsule
+///
+/// For general explanations about Capsules, see [`PyCapsule`].
+///
+/// This macro takes the following arguments:
+/// - segments of the full Python dotted name of the capsule
+/// - `$rustmod`, a suitable name to define a module
+/// - signature of the wished function (the macro will insert the `extern unsafe fn`)
```
markbt: As above, maybe:
```
/// This macro takes:
/// - The fully-dotted name of the module that contains the target capsule,
/// - The name of the target capsule,
/// - A rust module name, which will contain the retrieval function, and
/// - The signature of the function contained in the capsule.
```

src/objects/capsule.rs
```diff
@@ -0,0 +1,548 @@
...
+/// fn main() {
+///     create_capsule();
+///     retrieve_use_capsule();
+///     retrieve_use_capsule();
+/// }
+/// ```
+/// [`PyCapsule`]: struct.PyCapsule.html
+#[macro_export]
+macro_rules! py_capsule_fn {
+    ($($capsmod:ident).+, $capsname:ident, $rustmod:ident, $( $sig: tt)* ) => (
```
markbt: Again, maybe this could be:
```
   py_capsule_fn!(from full.path.to.module import functioncapsule as rustmod: fn(c_int) -> c_void)
```

src/objects/capsule.rs
```diff
@@ -0,0 +1,548 @@
...
+    ///   (although capsule data is usually static)
+    pub unsafe fn import_data<'a, T>(py: Python, name: &CStr) -> PyResult<&'a T> {
+        Ok(&*(Self::import(py, name)? as *const T))
+    }
+
+    /// Retrieves the contents of a capsule as a void pointer by its name.
+    ///
+    /// This is suitable in particular for later conversion as a function pointer
+    /// with `mem::transmute`, for architectures where data and function pointers have
+    /// the same size (see details about this the documentation of the Rust standard library).
```
markbt: Can we get a link here? 😃

src/objects/capsule.rs
```diff
@@ -0,0 +1,548 @@
...
+///     assert_eq!(capi.get_code(CStr::from_bytes_with_nul(b"COMMA\0").unwrap()), Ok(44));
+///     assert_eq!(capi.get_code(CStr::from_bytes_with_nul(b"\0").unwrap()),
+///                Err(UnicodeDataError::UnknownName));
+/// }
+/// ```
+/// [`PyCapsule`]: struct.PyCapsule.html
+/// [`PyCapsule::import_data`]: struct.PyCapsule.html#method.import_data
+#[macro_export]
+macro_rules! py_capsule {
+    ($($capsmod:ident).+, $capsname:ident, $rustmod:ident, $ruststruct: ident ) => (
```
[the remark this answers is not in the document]
gracinet: @markbt 

I like this suggestion a lot, it's indeed much better for readers to make it self-explanatory.
I'll give it a try.

src/objects/capsule.rs
```diff
@@ -0,0 +1,548 @@
...
+                        Ok(f) => Ok(*f),
+                        Err(ref e) => Err(e.clone_ref(py)),
+                    }
+                }
+            }
+        }
+    )
+}
+
+/// Capsules are the preferred way to export/import C APIs between extension modules,
```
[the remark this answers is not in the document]
gracinet: Done

src/objects/capsule.rs
gracinet: especially since it's `pub`, indeed. Done

gracinet:
@markbt thanks for the review ! For now, I'm making a new commit for each of the changes you requested, but I think, I'll rewrite the whole branch at the end in three commits (struct, data macro, fn macro) at the end of the process.

I'm getting rid of the trivial changes right now, but the macro changes will take me a longer while.

gracinet:
@markbt what do you think of the present state (especially the macro syntax) ?
Still need to rebase and squash in but 3 commits, also maybe add that the defaulting of the Rust struct to the capsule name.
I don't think there are many changes to be made for the macros declarations, because none of these is internal but I may be wrong on that part (hard to tell, with the current master not compiling yet with edition="2018")

gracinet:
@markbt 

I've been using this in both directions in the concrete case of Mercurial. This a for now in an unsubmitted topic yet but it works for me, and we'd need to rely on it quite soon now. 

The present macro syntaxes are:

```
py_capsule!(from some.py.mod import something as rustmod for RustStruct)
py_capsule_fn!(from some.py.mod import fun as rustmod signature (a: c_int) -> c_int)
```

I've made another branch for 2018 edition that shows that the macros don't need to be adapted, if you want to check it: https://github.com/gracinet/rust-cpython/tree/2018

gracinet:
A next step (probably a subsequent PR) would be to have default values for `RustStruct` (in the data case) and `rustmod` in the function case:

```
py_capsule!(from some.py.mod import something as rustmod)
py_capsule!(from some.py.mod import something as rustmod for RustStruct)
py_capsule!(from some.py.mod import something as rustmod fn(a: c_int) -> c_int)
```
with a 3-form `py_capsule!` macro deferring to two inner macros: `py_capsule_data!` and `py_capsule_fn!`

This would probably need some adaptation for the 2018 edition, although I'd plan to make the two helpers public, too (at least for backwards compatibility).

My take on this is that it can probably wait until `rust-cpython` switches to the 2018 edition. Perhaps my above mentionned branch can serve as a baseline work for this, too.

src/objects/capsule.rs
gracinet: Reformulated for the new macro syntax.

src/objects/capsule.rs
```diff
@@ -0,0 +1,548 @@
...
+///
+/// For more details on capsules, see [`PyCapsule`]
+///
+/// The caller has to define an appropriate `repr(C)` `struct` first, and put it in
+/// scope (`use`) if needed along the macro invocation.
+///
+/// The arguments of this macro are
+/// - the segments of the full dotted name of the targetted capsule,
+/// - `$rustmod`, a suitable module name,
+/// - `$ruststruct`, the name of the target `struct`.
```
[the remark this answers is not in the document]
gracinet: Reformulated for the new macro syntax

src/objects/capsule.rs
gracinet: Done with `signature` instead of `: fn`

PR #170, opened 2019-01-10
Use rustup.rs to install Rust in Travis builds

markbt:
The old Rust installer has been replaced with rustup.rs.  Use this to
install Rust in Travis.


markbt:
Rust nightly installer is broken today because of https://github.com/rust-lang/rust/issues/57488

PR #175, opened 2019-03-18
doc(readme) Fix the extension example

Hywan:
Thanks for the great project!

Without this patch, the example were not working on my computer with Python 3. I hope it's OK!

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ use cpython::{PyResult, Python};
 
 // add bindings to the generated python module
 // N.B: names: "librust2py" must be the name of the `.so` or `.pyd` file
-py_module_initializer!(librust2py, initlibrust2py, PyInit_librust2py, |py, m| {
+py_module_initializer!(librust2py, initlibrust2py, PyInit_rust2py, |py, m| {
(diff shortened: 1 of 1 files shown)


dgrunwald:
As the comment before that line says: "librust2py" must be the name of the `.so` or `.pyd` file
I think it's platform-dependent whether cargo prepends `lib`. Changing this line might fix it for Windows, but breaks it for Linux (and vice versa).

Hywan:
The library name starts with `lib`, but anyway, it's probably a macOS specific thing.

dgrunwald:
So Python prepends `lib` when searching for an extension module on macOS? `import name` -> `libname.dylib`? Though I believe it would also find `name.so` on MacOS.

I think this change is fine, having a `lib` in the init function name is unusual (would require `import libname` from Python). But the comment needs to be adjusted as well then, because extension modules on Windows are typically named `name.pyd` without a lib prefix.

Hywan:
The lib is correctly found, but Python tells me that the `PyInit_rust2py` must be declared, and _not_ `PyInit_librust2py`.

PR #177, opened 2019-03-23
Implement data marshalling module for python3-sys

indygreg:
I had a need for some of these APIs in a project. I went ahead and implemented the full `marshal.h` interface.


markbt:
Can this be done for Python 2.7, too, or is there a reason to exclude it?  It looks like the API is basically the same on both.

indygreg:
Yes, I can do 2.7 support. No good reason to exclude it. In my mind 2.7 is pretty much already dead, that's all.

PR #179, opened 2019-06-13
Docstrings in static and class methods

AdamRzepka:
https://github.com/dgrunwald/rust-cpython/pull/166 continued.

Support for docstrings in class methods and static methods. Slots (special methods) are still not supported and I don't see any way to implement it with Python C API.


markbt:
Thanks!

PR #180, opened 2019-06-15
Capsule: making ffi:PyObject available

gracinet:
This follow-up clarifies things for `PyCapsule` use cases where pointers to the C Python objects are
involved.

Some of the capsules from the standard library would probably benefit from this, my first use case would be the [make_dirstate_tuple() capsule of current Mercurial](https://www.mercurial-scm.org/repo/hg/rev/e240bec26626#l4.34)

It would be simpler and a bit more powerful to just add `pub type RawPyObject = ffi::PyObject` to `src/lib.rs`. I don't know if that's acceptable, but I could make a version based on this, also making `cpython::PyObject` API doc a bit less confusing.

### DIFF
--- a/src/objects/capsule.rs
+++ b/src/objects/capsule.rs
@@ -222,7 +222,13 @@ pyobject_newtype!(PyCapsule, PyCapsule_CheckExact, PyCapsule_Type);
 /// The `retrieve()` function is unsafe for the same reasons as [`PyCapsule::import_data`],
 /// upon which it relies.
 ///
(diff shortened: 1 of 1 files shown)


gracinet:
Had to fix again on 1.25 because of `extern crate` being inject inside the doctest generated `main()`

markbt:
I'm not sure that exposing `ffi::PyObject` in the general case would be a good idea (ideally we should hide these from the user).  In the case of Capsules I guess it's necessary, so restricting it to the capsule-generated module is probably a good thing.

PR #181, opened 2019-06-21
Version 0.3.0

markbt:


markbt:
I didn't get round to finishing off the 0.2.2 release.  With the change to the minimum Rust compiler version, I'm bumping to 0.3.0 instead.

PR #183, opened 2019-07-24
py_class: support Rust raw identifiers for method and keyword names

quark-zju:
Make it possible to use Rust keywords for keyword argument and method names.


markbt:
We probably need to bump the minimum Rust version to 1.30 as that's when raw identifiers were introduced.

PR #187, opened 2019-11-03
Support Python 3.8 in python3-sys.

dgrunwald:
This time around the diff between the Python 3.7.0 and Python 3.8.0 headers was non-trivial; because several parts of the non-limited API were moved around (mostly to the new internal `Include/cpython/*.h` headers).
So it's quite possible that I missed some relevant changes.

It turned out that rust-cpython mostly already works without these changes (enough that all our tests except for `check_symbols.py` pass on Python 3.8 without this PR!).
`PyCodeObject` is obviously broken without this PR, but it's unused in `rust-cpython`.
`PyTypeObject` is a bit more tricky, but in the usual case we're saved because we don't set `_Py_TPFLAGS_HAVE_VECTORCALL`, so an out-of-bounds read should not happen. Only in the rare case when compiling against a Python interpreter using the `COUNT_ALLOCS` mode it's important to have exactly the correct `PyTypeObject` definition.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,19 +4,13 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
+  - "3.7"
+  - "3.8"
 env:
   - RUST_VERSION=1.30.0
   - RUST_VERSION=nightly
 matrix:
   include:
-    - python: "3.7"
-      dist: xenial
-      sudo: true
-      env: RUST_VERSION=1.30.0
-    - python: "3.7"
-      dist: xenial
-      sudo: true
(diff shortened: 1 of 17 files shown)


dgrunwald:
Going forward we should consider using only the `PY_LIMITED_API` in the cpython crate. This would allow building extension modules where a single binary is compatible with multiple major Python versions.
A blocker here was that some important functionality like `PyIter_Check` was only available via macros that required the full API; but at least this particular case was finally fixed in Python 3.8.
I haven't checked yet if there are any other problematic cases.

PR #191, opened 2019-11-16
Define Python 3 exceptions from PEP 3151

indygreg:
According to https://docs.python.org/3/c-api/exceptions.html these were all defined in PEP 3151 and added in Python 3.3.


gracinet:
@indygreg nothing against adding missing exceptions on principle.

Don't hesitate to ping me if you see no progress in 2 weeks

gracinet:
Looks good and works for me, thanks,

A good thing that according to README, the minimal supported version of `rust-cpython` is 3.3, then :-) I did not go as far as to try with 3.3 though (noticed we had to drop it from Travis).

PR #192, opened 2019-11-16
Define inittab entities

indygreg:
The struct and function are documented at
https://docs.python.org/3/c-api/import.html#c._inittab
and defined in source code (in CPython master) at
https://github.com/python/cpython/blob/51edf8aaa2e17626f9690ed29d25945fc03016b9/Include/cpython/import.h#L30.

It is unclear whether PyImport_Inittab is part of the official
API, as it isn't clearly documented. But the static has been
shipping for years and is still in CPython master. I need to
use this variable in PyOxidizer in order to modify the set
of extensions at run-time. Putting it behind Py_LIMITED_API
should be safe.

### DIFF
--- a/python3-sys/src/import.rs
+++ b/python3-sys/src/import.rs
@@ -73,6 +73,14 @@ pub unsafe fn PyImport_ImportModuleEx(name: *const c_char,
 #[repr(C)]
 #[derive(Copy, Clone)]
 #[cfg(not(Py_LIMITED_API))]
+pub struct _inittab {
+    pub name: *mut c_char,
+    pub initfunc: Option<unsafe extern "C" fn()>,
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+#[cfg(not(Py_LIMITED_API))]
 pub struct _frozen {
     pub name: *const c_char,
     pub code: *const c_uchar,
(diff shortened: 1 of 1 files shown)


indygreg:
Are there plans to merge this?

python3-sys/src/import.rs
```diff
@@ -83,4 +91,7 @@ pub struct _frozen {
 #[cfg_attr(windows, link(name="pythonXY"))]
 extern "C" {
     pub static mut PyImport_FrozenModules: *const _frozen;
+    pub static mut PyImport_Inittab: *mut _inittab;
```
gracinet: If `_inittab` is conditioned by `Py_LIMITED_API` shouldn't this also?

gracinet:
Sorry for being slow, my bandwith being very constrained these days, I've been somewhat limited to straightforward stuff.

On principle, I'm fine with this and the `Py_LIMITED_API` (tied in turn to PEP 384 mode) feels relevant, yes.

Once the inconsistency in conditional flags is resolved (or turns out not to exist), I'm ready to merge this.

python3-sys/src/import.rs
indygreg: This whole block is protected by a `#[cfg(not(Py_LIMITED_API))]`.

PR #193, opened 2019-11-16
Support link mode control

indygreg:
As part of developing PyOxidizer, I needed to force python3-sys to
statically link against a Python library on Windows in a downstream
crate of python3-sys. This requires the unstable `static-nobundle`
link type so Cargo leaves symbols as unresolved when python3-sys
is built. (Currently, the `static` linkage type verifies referenced
symbols are present at crate build time.) See
https://github.com/rust-lang/rust/issues/37403 for more. Look for
comments by me (@indygreg) to describe the issue in more detail.

This commit teaches python3-sys a pair of new build features which
enable more explicit control over the linker directives emitted by
its build script. If no directive is specified, `link-mode-default`
is used and the existing logic for linker directive emission is
used. If `link-mode-unresolved-static` is used and we're on Windows,
we emit a `static-nobundle=pythonXY` linker directive and
omit the location of the library. This effectively says "I depend
on a static `pythonXY` library but don't resolve the symbols
when you build me and require someone else to specify the location
to that library." What PyOxidizer does is emit its own linker
directive that defines the location of a static `pythonXY` library,
satisfying the linker constraint and enabling the build to work.
If a downstream create doesn't do this, the build should fail due
to a missing library or symbols.

I have purposefully designed the crate features to be extensible.
If we want to add additional, mutually exclusive features in the
future, we could do that. e.g. we could add a `link-mode-static`
that force emits a `rustc-link-lib=static=pythonXY` directive
to force static linking, even if a shared library is detected.
But I have no need for this today and don't want to complicate
the code, so I haven't added it.

To round out the new feature, features have been added to the
cpython crate to toggle the new features.

Because Python 2.7 is end of life in a few weeks, I have not
implemented the new feature for Python 2.7. I suspect very
few people will use this feature anyway and I'm pretty confident
that nobody will request this feature on Python 2.7.

I concede that adding this feature to the crate to support
PyOxidizer's esoteric use case is a bit unfortunate. I really wish
Cargo allowed a crate to wholesale replace the build script output
of a dependency, as PyOxidizer could statically resolve the
Python settings for python3-sys since it brings its own Python
library. But Cargo doesn't have this feature. So I'm stuck
having to land this feature in the upstream crate to avoid having
to maintain a permanent fork of `rust-cpython`. Sorry :/

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,10 @@ extension-module = [ "python3-sys/extension-module" ]
 # or python3-sys. (honestly, we should probably merge both crates into 'python-sys')
 extension-module-2-7 = [ "python27-sys/extension-module" ]
 
+# Use these features to explicitly control linking for Python 3.
+# (See the documentation in python3-sys/Cargo.toml for more info.)
+py-link-mode-default = [ "python3-sys/link-mode-default" ]
+py-link-mode-unresolved-static = [ "python3-sys/link-mode-unresolved-static" ]
 
 # Optional features to support explicitly specifying python minor version.
 # If you don't care which minor version, just specify python3-sys as a 
--- a/python3-sys/Cargo.toml
+++ b/python3-sys/Cargo.toml
@@ -37,6 +37,30 @@ default = ["python-3"]
 # so that the module can also be used with statically linked python interpreters.
 extension-module = [ ]
 
+# This feature implies default linking behavior.
+#
+# If not an extension module or on Windows, the crate will link against
+# pythonXY where XY is derived from the discovered Python version. The link
+# type will be static, shared, or framework depending on the discovered Python.
+#
+# The path to pythonXY from the discovered Python install may also be
+# added to the linker search path.
+#
+# This link mode is used by default unless an alternate link mode feature is
+# used.
+link-mode-default = []
+
+# This feature forces Python symbols to be unresolved by emitting a
+# `rustc-link-lib=static-nobundle=pythonXY` directive on Windows (which
+# is the only platform where it makes sense).
+#
+# This mode is useful for scenarios where you want another crate to emit
+# the linker directives that define the location of a static Python library.
+#
+# This mode is typically not needed, as Python distributions on Windows
+# rarely use a static Python library.
+link-mode-unresolved-static = []
+
 # Bind to any python 3.x.
 python-3 = []
 
--- a/python3-sys/build.rs
+++ b/python3-sys/build.rs
@@ -316,16 +316,43 @@ fn configure_from_path(expected_version: &PythonVersion) -> Result<String, Strin
     let exec_prefix: &str = &lines[3];
 
     let is_extension_module = env::var_os("CARGO_FEATURE_EXTENSION_MODULE").is_some();
-    if !is_extension_module || cfg!(target_os="windows") {
-        println!("{}", get_rustc_link_lib(&interpreter_version,
-            ld_version, enable_shared == "1").unwrap());
-        if libpath != "None" {
-            println!("cargo:rustc-link-search=native={}", libpath);
-        } else if cfg!(target_os="windows") {
-            println!("cargo:rustc-link-search=native={}\\libs", exec_prefix);
-        }
+    let mut link_mode_default = env::var_os("CARGO_FEATURE_LINK_MODE_DEFAULT").is_some();
+    let link_mode_unresolved_static = env::var_os("CARGO_FEATURE_LINK_MODE_UNRESOLVED_STATIC").is_some();
+
+    if link_mode_default && link_mode_unresolved_static {
+        return Err("link-mode-default and link-mode-unresolved-static are mutually exclusive".to_owned());
     }
 
+    if !link_mode_default && !link_mode_unresolved_static {
+        link_mode_default = true;
+    }
+
+    if link_mode_default {
+        if !is_extension_module || cfg!(target_os="windows") {
+            println!("{}", get_rustc_link_lib(&interpreter_version,
+                ld_version, enable_shared == "1").unwrap());
+            if libpath != "None" {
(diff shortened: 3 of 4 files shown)


indygreg:
FWIW #191, #192, and this issue represent all the changes I've made to `rust-cpython` in order to support PyOxidizer (https://pyoxidizer.readthedocs.io/en/latest/). Once these changes are merged and a new `rust-cpython` release is made, I will be able to upload the `pyembed` crate to crates.io, as it will no longer require a forked version of `rust-cpython`. The `pyembed` crate provides a high-level API for controlling embedded Python interpreters and invoking build system magic to download and use pre-built Python binaries. This will make it vastly easier for Rust programmers to embed Python interpreters in Rust applications.

Merging these PRs isn't super urgent for me, as PyOxidizer has been getting by with a fork of `rust-cpython` for a while. But eliminating the `rust-cpython` fork will make the usability of PyOxidizer substantially better and I eagerly anticipate this improvement!

gracinet:
@indygreg thanks for the bunch of PRs. I agree with the general goal, and, needless to say, I wish `PyOxidizer` to succed. 

I've assigned a few to myself (those I felt confident with). I'll take a look at that one once I've reviewed those easier ones, though.

The fact that it's not super urgent aligns well with my present workload :-) 
Don't hesitate to ping me if you see no progress in two weeks time.


As you know, I don't have that much bandwith 

I will surely take a look at that one, but that should be after the easy ones. Don't hesitate to ping me after 2 weeks

indygreg:
It has been almost 2 weeks and I'm pinging @gracinet as requested. Also, add #194 to the set of issues required for PyOxidizer to not require a fork of this crate.

Cargo.toml
```diff
@@ -65,6 +65,10 @@ extension-module = [ "python3-sys/extension-module" ]
 # or python3-sys. (honestly, we should probably merge both crates into 'python-sys')
 extension-module-2-7 = [ "python27-sys/extension-module" ]
 
+# These these features to explicitly control linking for Python 3.
+# (See the documentation in python3-sys/Cargo.toml for more info.)
+link-mode3-default = [ "python3-sys/link-mode-default" ]
+link-mode3-unresolved-static = [ "python3-sys/link-mode-unresolved-static" ]
```
gracinet: Just nitpicking, but I don't feel the need to include this `3` in the new feature names. This would be consistent with  `extension-module`, that does not.

Maybe it'd be better to prefix with `python` or `py`, to make explicit it's about linking Python: `py-link-mode-default`, `py-link-mode-unresolved-static`?

python3-sys/Cargo.toml
```diff
@@ -37,6 +37,27 @@ default = ["python-3"]
...
 extension-module = [ ]
 
+# This feature implies default linking behavior.
+#
+# If not an extension module or on Windows, the crate will link against
+# pythonXY where XY is derived from the discovered Python version. The link
+# type will be static, shared, or framework depending on the discovered Python.
+#
+# The path to pythonXY from the discovered Python install may also be
+# added to the linker search path.
```
gracinet: Maybe here add to the comments that it's automatically activated (I can see that in `build.rs`) if other `link-mode` features aren't specified?

I suppose you didn't put it in `default` in order for the change to be transparently compatible for downstream crates that use `--no-default-features`, e.g, those that are for Python 2.7 ?

Cargo.toml
indygreg: I will incorporate this suggestion.

python3-sys/Cargo.toml
indygreg: Yes, I didn't put in in `default` because of implications for backwards compatibility or `--no-default-features`.

indygreg:
Regarding automated tests, I'm not sure how. We would have to inspect low-level `build.rs` output (how do you reliably do this from the context of a test?) or would need to build a Rust project providing its own link directive, which would be expensive. PyOxidizer exercises the functionality and I'll be sure to let you know if things regress :)

In the latest push, I also added an abort in the main crate's `build.rs` if in Python 2.7 mode and the link mode control features are present.

Cargo.toml
```diff
@@ -65,6 +65,10 @@ extension-module = [ "python3-sys/extension-module" ]
 # or python3-sys. (honestly, we should probably merge both crates into 'python-sys')
 extension-module-2-7 = [ "python27-sys/extension-module" ]
 
+# These these features to explicitly control linking for Python 3.
```
markbt: I assume this is a typo - `Use these features...`?

python3-sys/build.rs
```diff
@@ -316,16 +316,43 @@ fn configure_from_path(expected_version: &PythonVersion) -> Result<String, Strin
...
-        if libpath != "None" {
-            println!("cargo:rustc-link-search=native={}", libpath);
-        } else if cfg!(target_os="windows") {
-            println!("cargo:rustc-link-search=native={}\\libs", exec_prefix);
-        }
+    let mut link_mode_default = env::var_os("CARGO_FEATURE_LINK_MODE_DEFAULT").is_some();
+    let link_mode_unresolved_static = env::var_os("CARGO_FEATURE_LINK_MODE_UNRESOLVED_STATIC").is_some();
+
+    if link_mode_default && link_mode_unresolved_static {
+        return Err("link-mode-default and link-mode-unresolved-static are mutually exclusive".to_owned());
```
markbt: I think you have to be a little bit careful with mutually exclusive features, that's not really the way they were designed, as Cargo expects to be able to union them together when multiple crates depend on the same crate with different features.

That said, for this particular link mode case, it doesn't seem likely that would happen.

indygreg:
Thank you for accepting this! I was able to move PyOxidizer off a fork of `rust-cpython` and this makes me **very** happy! Don't hesitate to ping me if I need to clean up any mess my various PRs caused.

PR #194, opened 2019-11-16
Explicitly pass doc argument to py_method_def!

indygreg:
Without this, attempting to use py_fn! in an external crate
results in a compile error:

```
error: cannot find macro `py_method_def` in this scope
   --> pyembed/src/importer.rs:631:9
    |
631 | /         py_fn!(
632 | |             py,
633 | |             module_setup(
634 | |                 m: PyModule,
...   |
638 | |             )
639 | |         ),
    | |_________^
    |
    = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

Why this occurs, I'm not sure (I don't fully grok macros). But passing
an explicit value to the `doc` argument makes it work.

### DIFF
--- a/src/function.rs
+++ b/src/function.rs
@@ -149,7 +149,7 @@ macro_rules! py_fn_impl {
         }
         unsafe {
             $crate::_detail::py_fn_impl($py,
-                py_method_def!(_cpython__function__stringify!($f), 0, wrap))
+                py_method_def!(_cpython__function__stringify!($f), 0, wrap, ""))
         }
     }};
     // Form 2: inline function definition


gracinet:
Will jhave to take a close look at this. I confess I've never had to call in a `py_fn!` defined function from an external crate.

gracinet:
After some fiddling, I've been able to reproduce the issue at hand. That was a bit puzzling at first, because unless I got it wrong, using `py_fn!` from another crate is exactly what `tests/test_function.rs` does.

It seems to be actually a 2018 edition compatibility problem: put this `test_function.rs` in a new project (`cargo init`) under 2018 edition, remove the `#[macro_use]` in favor of `use cpython::py_fn`, and indeed, one gets the exact same error.

While this PR does resolve that particular problem, I'm not sure it's the best way to solve the issue. It seems just too coincidental.

@indygreg does simply adding `#[macro_use]` to your `extern crate cpython` work for you?

`

gracinet:
I wonder if the integration tests are always run with the same edition as the lib, and if we could have a systematic way to have some for these kinds of issues with macros.

Will have to dig in the Cargo book about this

gracinet:
The real cause is that `py_method_def!` calls itself if no docstring is supplied, but is not declared with `#[macro_export(local_inner_macros)]`.

I've made a new PR for that: #195 
@indygreg can you tell me if that works for you?

indygreg:
`#[macro_use] extern crate cpython` does fix the problem for me. The patch in #195 also fixes the problem without any changes to my source code. So I think I'll vendor the `function.rs` from #195 into my repository in anticipation of it getting merged and incorporated in the next release of this crate!

Let's close this PR in favor of your preferred solution.

PR #195, opened 2019-11-30
Edition 2018: made `py_method_def!` usable from external crates.

gracinet:
The first form macro is called by `py_fn!`, and calls `py_method_def!`
again with a default value for the docstring.

But it didn't bear the `local_inner_macros` modifier so that fails.
Adding that modifier forces us to rewrap `concat!` as done before
with `stringify!`.

Reference: https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#macros-using-local_inner_macros

### DIFF
--- a/src/function.rs
+++ b/src/function.rs
@@ -26,7 +26,7 @@ use conversion::ToPyObject;
 use ffi;
 use err::{self, PyResult};
 
-#[macro_export]
+#[macro_export(local_inner_macros)]
 #[doc(hidden)]
 macro_rules! py_method_def {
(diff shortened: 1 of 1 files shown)


gracinet:
This PR is the result of my review of #194, whose author acknowledges it works for him. It's best if someone else reviews it, but I would probably merge it myself as well if nobody can find the time.

markbt:
At some point we should just move to 2018 edition entirely.

gracinet:
Thanks!

PR #197, opened 2020-01-14
add append method to PyList

binh-vu:
This PR implements the append method for the PyList class because using the insert method is twice lower.

I also added test for this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "cpython"
-version = "0.3.0"
+version = "0.3.1"
 description = "Bindings to Python"
 authors = ["Daniel Grunwald <daniel@danielgrunwald.de>"]
 readme = "README.md"
--- a/src/objects/list.rs
+++ b/src/objects/list.rs
@@ -16,11 +16,11 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use python::{Python, PythonObject, ToPythonPointer, PyClone, PyDrop};
-use err::{self, PyErr, PyResult};
 use super::object::PyObject;
+use conversion::{FromPyObject, ToPyObject};
+use err::{self, PyErr, PyResult};
 use ffi::{self, Py_ssize_t};
-use conversion::{ToPyObject, FromPyObject};
+use python::{PyClone, PyDrop, Python, PythonObject, ToPythonPointer};
 
 /// Represents a Python `list`.
 pub struct PyList(PyObject);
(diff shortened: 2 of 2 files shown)


src/objects/list.rs
```diff
@@ -76,20 +80,29 @@ impl PyList {
         assert!(r == 0);
     }
 
+    /// Appends an item to the end of the list
+    pub fn append_item(&self, _py: Python, item: PyObject) {
```
gracinet: why not call it just `append`?  It would be consistent with the Python namings, and there's no ambiguity about what it does.

gracinet:
Thanks for the contribution, this looks quite reasonable to me, and it is after all only exposing a core Python method. Do you agree about the method name?

That being said, it would be nice if you could do that in two commits: one for the formatting, the other for the real change.

I believe you that inserting is slower, but if you have a reference for that, it would make a nice addition to the commit message as well. My opinion is that having a more straightforward, non-packing way to append on a list is worthwile on its own anyway.

gracinet:
fixed my comment, meant not panicking!

binh-vu:
Originally, I use `append`, but I see you are using `insert_item` instead of `insert`, so I want it to be consistent with you.

If you want to fix the function name, should we add another function called `insert` as well?

Will add the following benchmark between insert and append functions that I used to the next commit message.

```
from time import time

def a():
    L = []
    for i in range(10**5):
        L.append(i)
    return len(L)

def b():
    L = []
    for i in range(10**5):
        L.insert(len(L), i)
    return len(L)

start = time()
for i in range(5):
    a()
print(time() - start)

start = time()
for i in range(5):
    b()
print(time() - start)
```
Output:
```
0.037670135498046875 (append)
0.07758903503417969 (insert)
```

gracinet:
> Originally, I use `append`, but I see you are using `insert_item` instead of `insert`, so I want it to be consistent with you.

Ah yes, did not notice that, thanks for the explanation. I went through the history, but there's no explanation for that naming. It might have been just influence of `set_item`. 

> If you want to fix the function name, should we add another function called insert as well?

Nah, don't bother with that, I'm fine with `append_item`, we could add shorter aliases later on.
Please just add something about that in the commit message, so that future contributors can see your motivation was just consistency with `insert_item` and that's enough for me.

I don't think there's a collision with another use of `insert` in our context, but I don't feel it's worth the energy to really make sure of that.

gracinet:
> Will add the following benchmark between insert and append functions that I used to the next commit message.

Benchmarking from the Python side might yield different results than from native code, which is after all what rust-cpython crate users will call.

I'm just being curious here, as said earlier, I'm already happy with the simpler apend call

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 
 name = "cpython"
-version = "0.3.0"
+version = "0.3.1"
```
markbt: Is there a reason for this version bump?  We usually do that in a separate commit when we make the release (e.g. c7224ad2d 2e1ac24d6 b7090799b6)

markbt:
Thinking about it some more, I'm now in favour of renaming `insert_item` to `insert` (retaining the old one as a deprecated wrapper for now), which lets us call this `append`.

This is what I'd like to push to master if there are no objections: https://github.com/markbt/rust-cpython/tree/binh-vu-master

gracinet:
@markbt checked your branch, ran `cargo doc`, everything's fine: please go ahead

markbt:
Merged in 4432943

PR #199, opened 2020-01-24
improve PythonObjectDowncastError TypeErrors

markbt:
When TypeErrors are generated from PythonObjectDowncastErrors, include
the type that the downcast was targetted at, and the type of the object
that was being targetted.

This results in errors like:

    TypeError: Expected type that converts to PyString but received int


Alphare:
I am so happy that you did the work. I was getting mad with those "computer says no" errors.

PR #200, opened 2020-01-24
format the code

markbt:
Autoformatting the code makes it easier to make and merge changes.  I propose we run everything through rustfmt here.

I've made a few changes to make the formatting more stable.  For example, comments in the Rust AST are usually attached to the item that follows, so I've re-ordered the TODO items in the pythonX-sys crates.


Makefile
```diff
@@ -35,11 +35,13 @@ src/py_class/py_class_impl2.rs: src/py_class/py_class_impl.py
 src/py_class/py_class_impl3.rs: src/py_class/py_class_impl.py
 	PY=3 python $< >$@
 
-# To use this target: pip install cogapp
-cog: python27-sys/build.rs
-	cog.py -r $^
+python27-sys/build.rs: python3-sys/build.rs
+	@echo "// THIS FILE IS GENERATED FROM python3-sys/build.rs" > $@
+	@echo "// DO NOT MODIFY" >> $@
```
gracinet: Nice to have thought of that

python27-sys/build.rs
```diff
@@ -1,49 +1,42 @@
...
-//import os
-//import cog
-//HERE = os.path.dirname(cog.inFile)
-//for line in open(os.path.join(HERE, '../python3-sys/build.rs')):
-//    if line.strip():
-//        cog.out(line)
-//    else:
-//        cog.out("// ~~~~~~~~~~ generated file, modify `python3-sys/build.rs` ~~~~~~~~~~" + line)
-//]]]
+// THIS FILE IS GENERATED FROM python3-sys/build.rs
```
gracinet: Ah, it turns out not to be true in that case

gracinet: Ah sorry, I see it's on purpose actually.

markbt: I replaced the use of `cog` as it seemed superfluous for what it was doing, and the way the generated file annotations were inserted was causing conflicts with rustfmt.

gracinet:
I only skimmed through it quickly and checked cargo doc and reran the tests
Merging ASAP before it gets in conflict.

This is obviously useful, did you note the rustfmt version somewhere? 
I've been struggling so much with rustfmt changes in other projects myself…

PR #207, opened 2020-01-31
Implement reference extraction for slot functions and optional reference extraction

markbt:
This allows slot implementations with a single parameter (in particular,
`__compare__`, `__getitem__`, `__delitem__`, and in-place operators like `__iadd__`),
as well as `__richcmp__` to work with reference types (types that implement
RefFromPyObject) as well as full types (types that implement FromPyObject).

Additionally, optional references are included.  This is an improvement over
normal methods, so this is extended to normal methods, too.

The syntax is the same as for methods, e.g.

    def __contains__(&self, key: &KeyType) -> PyResult<bool>

will generate a method where the key will be borrowed from the PyObject,
rather than copied.

Similarly for optional references:

    def __contains__(&self, key: Option<&KeyType>) -> PyResult<bool>

will be the same, except that the key will be `None` if the user provided
Python `None`.

The `__setitem__` method only supports reference extraction for the first
parameter (the key).  The value still needs to be a fully extracted type.

Binary arithmetic operators, like `__add__`, are not included.

### DIFF
--- a/src/argparse.rs
+++ b/src/argparse.rs
@@ -152,7 +152,7 @@ pub fn parse_args(
 ///   If no type is specified, the parameter implicitly uses
 ///   `&PyObject` (format 1), `&PyTuple` (format 4) or `&PyDict` (format 6).
 ///   If a default value is specified, it must be a compile-time constant
-//    of type `ty`.
+///   of type `ty`.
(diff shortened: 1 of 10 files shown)


markbt:
This is now stacked on top of "modernize the codebase".

PR #208, opened 2020-02-03
Implement property support

markbt:
Add support for defining Python properties (attributes with getters and optional setters) in Rust.

Properties are defined inside `py_class!` with:

    @property def property_name(&self) -> PyResult<PropertyType> { ... }

where `property_name` is the name in Python, and `PropertyType` is any type that impls `ToPyObject`.

Property setters can optionally be defined as:

    @property_name.setter def set_property_name(
        &self,
        value: Option<PropertyType>
    ) -> PyResult<()> { ... }

If the value is `None` then the caller is deleting the property (`del obj.property_name`).

If the value is `Some(x)` then the caller is setting the property to `x`.

For setters, the `PropertyType` can be any type that impls `FromPyObject`, or a reference or optional reference to any type that impls `RefFromPyObject`.

In the latter case, the type of `value` is `Option<Option<&PropertyType>>`, where `None` means the value is being deleted, `Some(None)` means the property is being set to Python `None`, and `Some(Some(x))` means the property is being set to `x`.

### DIFF
--- a/src/argparse.rs
+++ b/src/argparse.rs
@@ -152,7 +152,7 @@ pub fn parse_args(
 ///   If no type is specified, the parameter implicitly uses
 ///   `&PyObject` (format 1), `&PyTuple` (format 4) or `&PyDict` (format 6).
 ///   If a default value is specified, it must be a compile-time constant
-//    of type `ty`.
+///   of type `ty`.
 ///  * `body`: expression of type `PyResult<_>`.
 ///     The extracted argument values are available in this scope.
 ///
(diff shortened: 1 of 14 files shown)


Alphare:
Looks like this needs a rebase now

markbt:
I've rebased both PRs.  It looks like @gracinet didn't spot my latest change that removed `std::mem::uninitialized` in favour of just initializing to zero.  I've included that here, but let me know if you want to drop it.

PR #210, opened 2020-02-24
Add more import APIs

indygreg:
I need some of these APIs to support in-memory extension module
importing in PyOxidizer. While I was adding the ones I needed, I
figured I would include some other missing symbols from
import.h as well.

### DIFF
--- a/python3-sys/src/import.rs
+++ b/python3-sys/src/import.rs
@@ -98,4 +98,24 @@ extern "C" {
     pub static mut PyImport_Inittab: *mut _inittab;
 
     pub fn PyImport_ExtendInittab(newtab: *const _inittab) -> c_int;
+
+    #[cfg(not(Py_3_7))]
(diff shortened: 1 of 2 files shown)


python3-sys/src/import.rs
```diff
@@ -98,4 +98,24 @@ extern "C" {
     pub static mut PyImport_Inittab: *mut _inittab;
 
     pub fn PyImport_ExtendInittab(newtab: *const _inittab) -> c_int;
+
+    #[cfg(not(Py_3_7))]
+    pub fn _PyImport_FindBuiltin(name: *const c_char) -> *mut PyObject;
+    #[cfg(Py_3_6)]
```
markbt: This should be `Py_3_7`, too.

indygreg: Good catch. I force pushed a fix.

PR #211, opened 2020-02-29
Implement PEP-587 initialization APIs

indygreg:
Python 3.8 implements several new APIs around interpreter configuration
and initialization as defined by PEP-587.

I will want to use many of these APIs in PyOxidizer.

This commit defines the new APIs.

The new APIs are split in CPython source between Include/initconfig.h
and Include/cpython/pylifecycle.h. All of the APIs are outside the
Py_LIMITED_API scope.

We didn't have a pylifecycle.rs for the existing Include/pylifecycle.h.
Assuming we don't want to have different Rust modules distinguish between
the Include/<foo>.h and Include/cpython/<foo>.h variants (which appear
to have been introduced in CPython 3.8 - to help split Py_LIMITED_API I
think), I went ahead and created pylifecycle.rs. Ideally the Rust
modules would reflect the current state of the symbols in the latest
CPython sources. But I left this as a TODO.

To be honest, given the number of introduced symbols, I will be shocked
if there isn't a typo or bug lingering in here somewhere. I attempted
to port every symbol from Include/initconfig.h and
Include/cpython/pylifecycle.h. I attempted to preserve pointer
const/mut: if the source didn't use const, I didn't either. Although I
initially got some of this wrong during implementation and this is an
area that deserves extra review scrutiny.

### DIFF
--- a/python3-sys/src/initconfig.rs
+++ b/python3-sys/src/initconfig.rs
@@ -1,0 +1,168 @@
+// This entire module is Python 3.8+ and !Py_LIMITED_API only.
+
+use crate::pyport::Py_ssize_t;
+use libc::{c_char, c_int, c_ulong, wchar_t};
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub enum PyStatusType {
+    _PyStatus_TYPE_OK = 0,
+    _PyStatus_TYPE_ERROR = 1,
+    _PyStatus_TYPE_EXIT = 2,
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct PyStatus {
+    _type: PyStatusType,
+    func: *const c_char,
+    err_msg: *const c_char,
+    exitcode: c_int,
+}
+
+#[cfg_attr(windows, link(name = "pythonXY"))]
+extern "C" {
+    pub fn PyStatus_Ok() -> PyStatus;
+    pub fn PyStatus_Error(err_msg: *const c_char) -> PyStatus;
+    pub fn PyStatus_NoMemory() -> PyStatus;
+    pub fn PyStatus_Exit(exitcode: c_int) -> PyStatus;
+
+    pub fn PyStatus_IsError(err: PyStatus) -> c_int;
+    pub fn PyStatus_IsExit(err: PyStatus) -> c_int;
+    pub fn PyStatus_Exception(err: PyStatus) -> c_int;
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct PyWideStringList {
+    length: Py_ssize_t,
+    items: *mut *mut wchar_t,
+}
+
+#[cfg_attr(windows, link(name = "pythonXY"))]
+extern "C" {
+    pub fn PyWideStringList_Append(list: *mut PyWideStringList, item: *const wchar_t) -> PyStatus;
+    pub fn PyWideStringList_Insert(
+        list: *mut PyWideStringList,
+        index: Py_ssize_t,
+        item: *const wchar_t,
+    ) -> PyStatus;
+}
+
+#[repr(C)]
+#[derive(Clone)]
+pub struct PyPreConfig {
+    _config_init: c_int,
+    parse_argv: c_int,
+    isolated: c_int,
+    use_environment: c_int,
+    configure_locale: c_int,
+    coerce_c_locale: c_int,
+    coerce_c_locale_warn: c_int,
+    #[cfg(windows)]
+    legacy_windows_fs_encoding: c_int,
+    utf8_mode: c_int,
+    dev_mode: c_int,
+    allocator: c_int,
+}
+
+#[cfg_attr(windows, link(name = "pythonXY"))]
+extern "C" {
+    pub fn PyPreConfig_InitPythonConfig(config: *mut PyPreConfig) -> ();
+    pub fn PyPreConfig_InitIsolatedConfig(config: *mut PyPreConfig) -> ();
+}
+
+#[repr(C)]
+#[derive(Clone)]
+pub struct PyConfig {
+    _config_init: c_int,
+    isolated: c_int,
+    use_environment: c_int,
+    dev_mode: c_int,
+    install_signal_handlers: c_int,
+    use_hash_seed: c_int,
+    hash_seed: c_ulong,
+    faulthandler: c_int,
+    tracemalloc: c_int,
+    import_time: c_int,
+    show_ref_count: c_int,
+    show_alloc_count: c_int,
+    dump_refs: c_int,
+    malloc_stats: c_int,
+    filesystem_encoding: *mut wchar_t,
+    filesystem_errors: *mut wchar_t,
+    pycache_prefix: *mut wchar_t,
+    parse_argv: c_int,
+    argv: PyWideStringList,
+    program_name: *mut wchar_t,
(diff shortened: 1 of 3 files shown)


indygreg:
I just wanted to gently nudge a review of this PR. Having access to these new APIs is softly blocking PyOxidizer from utilizing Python 3.8 to its fullest extent.

While I'm here, PyOxidizer is currently depending on an unreleased version of rust-cpython. This means I can't release a new version until rust-cpython is released. I'm still a few days of hacking away from a new PyOxidizer release though. So don't let me pressure you. I will suggest that given the change velocity of this crate, perhaps it is worth performing a release every time any small batch of commits land?

gracinet:
Starting to review now. 

About the possibility of typos, I wonder if we shouldn't try and use [bindgen](https://crates.io/crates/bindgen) at some point.

python3-sys/src/initconfig.rs
```diff
@@ -0,0 +1,143 @@
...
+
+#[cfg_attr(windows, link(name = "pythonXY"))]
+extern "C" {
+    pub fn PyWideStringList_Append(list: *mut PyWideStringList, item: *const wchar_t) -> PyStatus;
+    pub fn PyWideStringList_Insert(list: *mut PyWideStringList, index: Py_ssize_t, item: *const wchar_t) -> PyStatus;
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct PyPreConfig {
```
gracinet: (detail) this `PyPreConfig` and `PyConfig` seem to be used only through pointers. Do they really need to be `Copy`, then?

python3-sys/src/lib.rs
```diff
@@ -48,6 +50,7 @@ pub use crate::pydebug::*;
 pub use crate::pyerrors::*;
 #[cfg(Py_3_4)]
 pub use crate::pyhash::*;
+pub use crate::pylifecycle::*;
```
gracinet: (detail) Since all the statements in `pylifecycle` are gated through `#[cfg(Py_3_8)]` wouldn't it be more consistent to just put that flag here? Same for `not(Py_LIMITED_API)` unless you think some of these might make their way to the limited API (although PEP 384 says it should be `pyconfig.h` and `Python.h` only)

python3-sys/src/pylifecycle.rs
```diff
@@ -0,0 +1,50 @@
...
+#[cfg_attr(windows, link(name = "pythonXY"))]
+extern "C" {
+    pub fn Py_PreInitialize(src_config: *const PyPreConfig) -> PyStatus;
+    pub fn Py_PreInitializeFromBytesArgs(src_config: *const PyPreConfig, argc: Py_ssize_t, argv: *mut *mut c_char) -> PyStatus;
+    pub fn Py_PreInitializeFromArgs(src_config: *const PyPreConfig, argc: Py_ssize_t, argv: *mut *mut wchar_t) -> PyStatus;
+
+    pub fn _Py_IsCoreInitialized() -> c_int;
+
+    pub fn Py_InitializeFromConfig(config: *const PyConfig) -> PyStatus;
+    pub fn _Py_InitializeMain() -> PyStatus;
```
gracinet: (detail) I suppose you omitted `_Py_InitializeFromArgs` and `_Py_InitializeFromWideArgs` because they have been removed for the future Python 3.9 (that's at least what I see in 3.9.0a4). Then a comment about that would be nice.

python3-sys/src/initconfig.rs
```diff
@@ -0,0 +1,143 @@
...
+#[cfg_attr(windows, link(name = "pythonXY"))]
+extern "C" {
+    pub fn PyConfig_InitPythonConfig(config: *mut PyConfig) -> ();
+    pub fn PyConfig_InitIsolatedConfig(config: *mut PyConfig) -> ();
+    pub fn PyConfig_Clear(config: *mut PyConfig) -> ();
+    pub fn PyConfig_SetString(config: *mut PyConfig, config_str: *mut *mut wchar_t, value: *const wchar_t) -> PyStatus;
+    pub fn PyConfig_SetBytesString(config: *mut PyConfig, config_str: *mut *mut wchar_t, value: *const c_char) -> PyStatus;
+    pub fn PyConfig_Read(config: *mut PyConfig) -> PyStatus;
+    pub fn PyConfig_SetBytesArgv(config: *mut PyConfig, argc: Py_ssize_t, argv: *mut *const c_char) -> PyStatus;
+    pub fn PyConfig_SetArgv(config: *mut PyConfig, argc: Py_ssize_t, argv: *mut *const wchar_t) -> PyStatus;
```
gracinet: In `initconfig.h`, argv in these two are `* const *`. I believe this translates as `*const *mut`

gracinet:
I'm rather agnostic with respect to the file organisation, although I must confess I didn't look up why it's been split that way in CPython. If there are other cases like that, we may want to reproduce the same structure for easier comparison, but it's not really important to me.

I went over the whole, except for the `char * const* argv` thing, I only have formal, minor, remarks.

gracinet:
with respect to releases, I agree we should have one soon, but there is one thing that I'd like to tackle first, and I am releasing another unrelated project tomorrow.

gracinet:
Forgot to mention: rust-cpython is now formatted with rustfmt. Could you please run it as well?

python3-sys/src/initconfig.rs
indygreg: Nope. I'm not even sure why I added `Copy`. Maybe I was basing it on patterns I saw elsewhere? I generally like avoiding `Copy` unless absolutely necessary because the implicit nature of `Copy` over `.clone()` can get you in trouble. I'll remove `Copy`.

python3-sys/src/lib.rs
indygreg: The reason I didn't gate all of `pylifecycle` is because there are symbols like `Py_Initialize` that are actually defined in `pylifecycle.h` but are currently defined elsewhere in `python3-sys`. I think the symbols were moved in Python 3.8. I have created the Rust `pylifecycle` module with the intent of eventually holding these symbols.

python3-sys/src/pylifecycle.rs
indygreg: More or less. They were removed from 3.9. But the reason they were removed was because they were declared in the `.h` but had no implementation! These symbols caused linking failures on CI due to missing symbols!

python3-sys/src/initconfig.rs
indygreg: You are correct. Will change.

indygreg:
Force pushed with changes to all points of feedback, including running `rustfmt`.

Regarding `bindgen`, the thought had crossed my mind. There's certainly been enough churn in the CPython internals to possibly warrant it. Although I'm unsure how much effort it would entail, as I'm unsure how `bindgen` could possibly deal with things like having multiple versions of the Python API. Perhaps we could use `bindgen` to generate skeleton bindings then apply a bunch of transformations to normalize them to the state we need. I dunno.

gracinet:
Can't find the my comment about gating `pylifecycle.rs` again, but anyway, if the plan is to introduce symbols that aren't thus restricted, now I understand it.

gracinet:
And thanks for the contribution!

python3-sys/src/lib.rs
gracinet: found it ! Yes, that makes sense.

PR #216, opened 2020-04-08
Travis CI: also testing on Rust stable

gracinet:
Since there's a good difference between our minimal supported Rust version
and the current stable, I feel it's worthwile to add a case with Rust
stable.


gracinet:
So, this reproduces the build failure first noticed in #215

PR #217, opened 2020-04-19
Make fields of initialization config public and implement Default

indygreg:
The structs added in #211 did not have public fields. The
fields need to be public in order to allow modifying them.

Furthermore, instantiating new instances of these structs would
be cumbersome (even with public fields) because there are so
many fields. So, we implement Default to construct new instances
that are zero initialized.

### DIFF
--- a/python3-sys/src/initconfig.rs
+++ b/python3-sys/src/initconfig.rs
@@ -14,10 +14,23 @@ pub enum PyStatusType {
 #[repr(C)]
 #[derive(Copy, Clone)]
 pub struct PyStatus {
-    _type: PyStatusType,
-    func: *const c_char,
-    err_msg: *const c_char,
-    exitcode: c_int,
+    pub _type: PyStatusType,
+    pub func: *const c_char,
+    pub err_msg: *const c_char,
+    pub exitcode: c_int,
+}
+
+impl Default for PyStatus {
+    fn default() -> Self {
+        // zeroed() is UB for enums. So we are explicit about
+        // what value it is set to. This probably isn't necessary
+        // as we are dealing with a C-compat struct and 0 is a valid
+        // enum value. But explicit is better than UB.
+        let mut status: Self = unsafe { core::mem::zeroed() };
+        status._type = PyStatusType::_PyStatus_TYPE_OK;
+
+        status
+    }
 }
 
 #[cfg_attr(windows, link(name = "pythonXY"))]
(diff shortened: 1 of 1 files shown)


python3-sys/src/initconfig.rs
```diff
@@ -14,10 +14,16 @@ pub enum PyStatusType {
...
-    exitcode: c_int,
+    pub _type: PyStatusType,
+    pub func: *const c_char,
+    pub err_msg: *const c_char,
+    pub exitcode: c_int,
+}
+
+impl Default for PyStatus {
+    fn default() -> Self {
+        unsafe { core::mem::zeroed() }
```
gracinet, line 25: Since `PyStatus` is an `enum`, and enums are prime examples of UB in the [doc behind zeroed()](https://doc.rust-lang.org/nightly/core/mem/union.MaybeUninit.html#method.zeroed), I think this warrants at least a comment. I think in this case it's ok, because 0 is indeed among the target values in the enum definition.

Also maybe, mention that `NULL` is an admissible value for these pointers on the Python side (I understand we're in FFI unsafe territory anyway).

indygreg: I changed the code here to avoid UB.

As for whether `Default` is needed, the typical pattern here is:

```
let mut pre_config = pyffi::PyPreConfig::default();
unsafe {
    pyffi::PyPreConfig_InitIsolatedConfig(&mut pre_config);
}
```

You need to obtain a struct instance to pass to a Python API which will "initialize" it. It doesn't really matter what the initial values are, since the Python APIs populate all fields AFAICT.

The Python APIs simply set default values for a given profile. It will be common for callers to want to update those fields before the struct is used to initialize the interpreter.

markbt: The way I read the docs for `zeroed()`, it's not UB if 0 is a valid value (which it is, because the enum has a `0` value and it's `repr(C)` for good measure).

That said, I don't think there's any harm to adding the extra set.  The compiler ought to elide it anyway.

gracinet: @markbt, that's what I thought, too, yes.

indygreg:
Is anything more needed to land this? I don't have a strong preference on use/non-use of `zeroed()`.

Typically after a struct instance is created, one calls a function like https://docs.python.org/3/c-api/init_config.html#c.PyPreConfig_InitPythonConfig to populate its fields. See https://github.com/indygreg/PyOxidizer/blob/48a4cab593a53a1c11bf313981f6b29818941ee0/pyembed/src/interpreter_config.rs#L253 and https://github.com/indygreg/PyOxidizer/blob/48a4cab593a53a1c11bf313981f6b29818941ee0/pyembed/src/interpreter_config.rs#L304 for what this might look like in Rust. I believe the CPython functions will set all fields to default values so there is no UB at play. But I'm not 100% and `zeroed()` feels safer to me.

markbt:
I think that's fine.  The setting of the enum value is probably superfluous, as they're `repr(C)`, but I won't let that stop it.

Issue #219, opened 2020-04-22
py_fn fails if a trailing comma is present


cpterry:
The `py_fn` macro fails if a trailing comma is present after the last argument of the function. This comma is added by `rustfmt` if those arguments are split across multiple lines so it's troublesome to work around.

An example of the code that will not compile is below. The only change from the README.md example is adding the comma in `b: i64,`.

```rust
use cpython::{py_fn, py_module_initializer, PyResult, Python};

// add bindings to the generated python module
// N.B: names: "rust2py" must be the name of the `.so` or `.pyd` file
py_module_initializer!(rust2py, |py, m| {
    m.add(py, "__doc__", "This module is implemented in Rust.")?;
    m.add(
        py,
        "sum_as_string",
        py_fn!(py, sum_as_string_py(a: i64, b: i64,)),
    )?;
    Ok(())
});

// logic implemented as a normal rust function
fn sum_as_string(a: i64, b: i64) -> String {
    format!("{}", a + b).to_string()
}

// rust-cpython aware function. All of our python interface could be
// declared in a separate module.
// Note that the py_fn!() macro automatically converts the arguments from
// Python objects to Rust values; and the Rust return value back into a Python object.
fn sum_as_string_py(_: Python, a: i64, b: i64) -> PyResult<String> {
    let out = sum_as_string(a, b);
    Ok(out)
}

```

lausek:
I've extended the macro responsible for parsing the parameter list to ignore automatically appended commas. Could you please check if that fixes your error?

cpterry:
That fixed it! Perfect! Thanks.

PR #220, opened 2020-04-23
fix for #219: ignore trailing comma in plist parsing

lausek:


markbt:
Looks ok to me.  Will let others comment before I merge it, in case I am missing something.

Thanks!

markbt:
Merged

Issue #221, opened 2020-04-24
No mappingproxy implementation?


e0gs:
I'm trying to check class attributes with Class.__dict__ with python3, and there seems to be nothing I can do with the returned mappingproxy.

It can't be cast into a PyDict, using call_method "items" or "keys" returns dict_items, dict_keys which in turn cannot be cast into PyDict, PyList or PySequence.

Even calling Class.__dict__.__iter__() and trying cast into PyIterator didn't work.

PR #223, opened 2020-04-25
add PyDictProxy, see #221

e0gs:
The mappingproxy functions are defined as static in https://github.com/python/cpython/blob/3.7/Objects/descrobject.c#L806 and tbh, dunno what's the point is with _Py_IDENTIFIER(keys) and _PyObject_CallMethodId for example in mappingproxy_keys so went directly to the source in abstracts by checking _Py_IDENTIFIER(keys) in https://github.com/python/cpython/blob/088b63ea7a8331a3e34bc93c3b873c60354b4fad/Tools/c-analyzer/known.tsv#L1212 etc.. https://github.com/python/cpython/blob/4a21e57fe55076c77b0ee454e1994ca544d09dc0/Objects/abstract.c#L2303

Anyway, everything works as expected and so on.

### DIFF
--- a/python3-sys/src/descrobject.rs
+++ b/python3-sys/src/descrobject.rs
@@ -1,8 +1,9 @@
 use libc::{c_char, c_int, c_void};
 
 use crate::methodobject::PyMethodDef;
-use crate::object::{PyObject, PyTypeObject};
+use crate::object::*;
 use crate::structmember::PyMemberDef;
+use crate::pyport::Py_ssize_t;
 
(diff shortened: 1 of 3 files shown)


markbt:
I think I understand this, but could you add some more comments in the commit message (and ideally improve the commit title, too, so that it fits in more with the other commits in the repo).

Python 2.7 is still a supported platform for now (e.g. Mercurial uses it).  Can you make the corresponding changes in python27-sys, please?

PR #227, opened 2020-05-18
WIP: Add support for the buffer protocol

### LINKED ISSUE #190: create PyBuffer from rust code
So I've had experience before creating python bindings with the C++ library [pybind11](https://github.com/pybind/pybind11) where I'm able to return python buffers but I can't figure out how to do the same with rust-cpython

mloebel:
This adds initial support for implementing the [buffer protocol](https://docs.python.org/3/c-api/buffer.html) via `py_class! {}`.

This works by defining a magic method:

```python
def __buffer__(&self) -> PyResult<MyBuffer> { ... }
```

where `MyBuffer` has to implement the following trait.

```rust
pub unsafe trait BufferHandle: 'static + Send {
    /// Returns the data of `Self` as a continous array of bytes.
    ///
    /// # Safety
    ///
    /// This needs to return an address that remains valid
    /// and stable if `self` gets moved or transformed to or from a void pointer.
    fn as_bytes(&self) -> &[u8];

    /// Convert `self` into an owned void pointer.
    ///
    /// # Safety
    ///
    /// The returned pointer has to be a valid pointer that
    /// can be converted back to `Self`.
    fn to_owned_void_pointer(self) -> *mut libc::c_void;

    /// Convert an owned void pointer back to `Self`. This takes owenrship of the pointer.
    ///
    /// # Safety
    ///
    /// The passed `ptr` has been created by this trait, and is only
    /// used at most once to convert back to `Self`.
    unsafe fn from_owned_void_pointer(ptr: *mut libc::c_void) -> Self;
}
```

## Remaining issues

- Does not work for Python2.7 for reasons I don't understand yet. The support for the "new" python 3 buffer interface exists in 2.7 with all the right type definitions, but the tests fails at runtime with "cannot make memory view because object does not have the buffer interface", so I've disabled it there for now. Help would be appreciated with this, as I'm not actively using 2.7 (which is eol, besides 😅).
- Only supports exporting a read-only byte array view for now.
- The `BufferHandle` API does not allow rejecting a buffer request.
- The `BufferHandle` API does not allow keeping ownership of the memory inside the class, as I was not sure if that can be safely done.

---

This partially fixes #190

### DIFF
--- a/python3-sys/src/memoryobject.rs
+++ b/python3-sys/src/memoryobject.rs
@@ -21,6 +21,7 @@ extern "C" {
         size: Py_ssize_t,
         flags: c_int,
     ) -> *mut PyObject;
+    pub fn PyMemoryView_FromBuffer(view: *mut Py_buffer) -> *mut PyObject;
     pub fn PyMemoryView_GetContiguous(
         base: *mut PyObject,
         buffertype: c_int,
--- a/python3-sys/src/object.rs
+++ b/python3-sys/src/object.rs
@@ -109,13 +109,13 @@ mod bufferinfo {
         }
     }
 
-    pub type getbufferproc = extern "C" fn(
+    pub type getbufferproc = unsafe extern "C" fn(
         arg1: *mut crate::object::PyObject,
         arg2: *mut Py_buffer,
         arg3: c_int,
     ) -> c_int;
     pub type releasebufferproc =
-        extern "C" fn(arg1: *mut crate::object::PyObject, arg2: *mut Py_buffer) -> ();
+        unsafe extern "C" fn(arg1: *mut crate::object::PyObject, arg2: *mut Py_buffer) -> ();
 
     /// Maximum number of dimensions
     pub const PyBUF_MAX_NDIM: c_int = 64;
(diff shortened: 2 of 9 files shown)


mloebel:
Update: Turns out that some of my use-cases for the buffer API use [`PyArg`s weird definition](https://docs.python.org/3/c-api/arg.html#strings-and-buffers) of what a "read-only bytes-like" object is: Something without a destructor for created `PyBuffer`s.

In practice this means that the buffer memory exported by the object has to be manged by the exporting object itself, and needs to be immutable for the entire lifetime of the object. 

I tried to model this by adding an alternative magic method that expects this signature: 
```rust
for<'a> fn __direct_buffer__(&'a Self, Python<'a>) -> PyResult<&'a [u8]>;
```

...with the idea that you can only implement this by returning references to `data` members or static memory. As far as I understood, `data` members are allocated in a stable memory location as part of the `PyObject` created by `py_class!`, so retrieving and accessing the `&'a [u8]` should be valid at any point as long as the object is alive. And by requiring it to be a direct reference to self, we prevent mutation through internal mutability wrapper.

PR #243, opened 2020-10-15
Add Python 3.9 support

dgrunwald:
As usual with new Python releases, I diffed the 3.8.0 Python headers against the 3.9.0 Python headers and manually made these adjustments to python3-sys.

While rust-cpython could be compiled against Python 3.9 without these changes, there were some breaking changes to the headers:
 * some macros like `PyObject_IS_GC` or `PyObject_GET_WEAKREFS_LISTPTR` were replaced with functions
 * some functions were removed from the API
 * `PyConfig` had breaking changes to the struct layout

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
   - "3.6"
   - "3.7"
   - "3.8"
+  - "3.9"
 env:
   - RUST_VERSION=1.32.0
   - RUST_VERSION=stable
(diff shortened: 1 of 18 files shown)


markbt:
I haven't compared with the Python headers, but it all looks reasonable to me.

The Travis CI is failing because Python 3.9 isn't available there yet.  Is this likely to become available soon, or are we going to need to do something else?

dgrunwald:
No idea when travis will add 3.9. See https://travis-ci.community/t/python-3-9-0-build/10091/10
We could switch to 3.9-dev in the meantime, though I'm not sure which version exactly we'd be getting there.

dgrunwald:
I've reported the ICE with nightly rust at https://github.com/rust-lang/rust/issues/78398

PR #244, opened 2020-10-30
Avoid abort() if the Python function never returns (thread exit)

quark-zju:
The AbortOnDrop is used in this pattern:

    let guard = AbortOnDrop(...);
    call_into_python(...);
    mem::forget(guard);

There is a special case that call_into_python does not return
(ex. the thread gets dropped). If that happens with pthread backend,
the `mem::forget` does not run, but `guard::drop` still runs, causing
the program to abort.

Remove AbortOnDrop to make that multi-thread case work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ paste = "0.1"
 serde = { version = "1", features = ["derive"], optional = true }
 
 [dev-dependencies]
+rustversion = "1.0"
 serde_bytes = { version = "0.11" }
 serde_cbor = { version = "0.11" }
 
--- a/Makefile
+++ b/Makefile
@@ -11,11 +11,13 @@ NIGHTLY := 0
 endif
 endif
 
+FEATURES := serde-convert
+
 ifeq ($(PY),2)
-FEATURES := python27-sys
+FEATURES := $(FEATURES) python27-sys
 endif
 ifeq ($(PY),3)
-FEATURES := python3-sys
+FEATURES := $(FEATURES) python3-sys
 ifdef PEP384
 export PEP384=1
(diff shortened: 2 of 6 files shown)


dgrunwald:
AbortOnDrop is critical to prevent Rust panics from causing undefined behavior; we can't simple delete those.

quark-zju:
Is there an example UB that is prevented by `AbortOnDrop`? It seems to me that `panic::catch_unwind` would capture some panic cases, and the rest of the panic cases would abort the program, so there is no need to abort again.

dgrunwald:
I don't think we're using `catch_unwind` everywhere you deleted `AbortOnDrop`.

Is there some way we could tell the difference between a panic and a thread exit?

quark-zju:
It seems `pthread_cleanup_push` might be used for pthread implementation. But I'm not sure about all platforms that Rust and Python supports. I think it's more portable to just avoid `AbortOnDrop` if `catch_unwind` is used. I'll update the code.

quark-zju:
CI is finally passing. It turns out that `catch_unwind` on Rust < 1.40 is buggy. `pthread_exit` throws C++ exception `abi::___forced_unwind` and gets silenced by `catch_unwind`. pthread then aborts with `FATAL: exception
not rethrown`. I added comments about details in the test. I didn't bump the minimal required rustc version since affected programs will abort both before and after the change.

dgrunwald:
This could still cause undefined behavior if someone adds a custom `CallbackConverter` impl that panics inside `error_value()`.

I guess we could keep the `AbortOnDrop`, but move it into the error handling code path, so that's it's not on the stack during the `catch_panic`.

PR #247, opened 2020-12-21
`PyString::data()`: return the internal representation of the Python unicode object

### LINKED ISSUE #246: PyString::to_string panics on UnicodeDecodeError
[`PyString::to_string`] returns a `PyResult<Cow<str>>`, so Python's raising an exception should cause [`PyString::to_string`] to return `Err(PyErr)` holding a `UnicodeDecodeError`. However, [`PyString::to_string`] panics if a Python `str` holds an unpaired surrogate, e.g.,
```python
"foo \ud800 bar"
```
The problem arises because [`PyString::to_string`]'s implementation relies on [`PyString::data`], whose API is fundamentally incompatible: [`PyString::data`] does _not_ return a [`PyResult`]. Thus, when the call to [`PyUnicode_AsUTF8AndSize`] in [`PyString::data`] raises an exception, the exception is not handled: https://github.com/dgrunwald/rust-cpython/blob/633dafc2c9e0ec22f8ff2479dd1cdadeb4ac0f40/src/objects/string.rs#L291-L295

The fix, I believe, is to
- [ ] call [`PyUnicode_AsUTF8AndSize`] directly in [`PyString::to_string`]'s implementation and handle its exception rather than relying on [`PyString::data`]
  - Implementation note: Code can be shared between [`PyString::to_string`] and [`PyString::data`], with the latter `unwrap()`ing and the former returning the [`PyResult`]
  - Alternative: Implement PyO3's [`PyString::to_str(&self, py: Python) -> PyResult<&str>`], which [does not have the panic bug](https://github.com/PyO3/pyo3/blob/b444f985868eef44dfe720a3872afabbbbf4a037/src/types/string.rs#L41-L68)
- [ ] Document that [`PyString::data`] may panic if the Python `str` is not representable as a Rust string
- [ ] Add a test case for this (e.g., [PyO3's test case])

[`PyString::to_string`]: http://dgrunwald.github.io/rust-cpython/doc/cpython/struct.PyString.html#method.to_string
[`PyResult`]: http://dgrunwald.github.io/rust-cpython/doc/cpython/type.PyResult.html
[`Cow`]: https://doc.rust-lang.org/std/borrow/enum.Cow.html
[`str`]: https://doc.rust-lang.org/std/primitive.str.html
[`PyErr`]: http://dgrunwald.github.io/rust-cpython/doc/cpython/struct.PyErr.html
[`PyUnicode_AsUTF8AndSize`]: https://docs.python.org/3/c-api/unicode.html?highlight=exception#c.PyUnicode_AsUTF8AndSize
[`PyString::data`]: http://dgrunwald.github.io/rust-cpython/doc/cpython/struct.PyString.html#method.data
[`PyString::to_str(&self, py: Python) -> PyResult<&str>`]: https://docs.rs/pyo3/0.12.4/pyo3/types/struct.PyString.html#method.to_str
[PyO3's test case]: https://github.com/PyO3/pyo3/blob/b444f985868eef44dfe720a3872afabbbbf4a037/src/types/string.rs#L261-L268

dgrunwald:
This fixes #246: panic in `PyString::to_string` and `PyString::to_string_lossy` when a Python3 unicode string contains unpaired surrogates.

As an optimization, `PyString::to_string` keeps using `PyUnicode_AsUTF8AndSize` which allows Python to cache the UTF-8 representation.

### DIFF
--- a/src/objects/string.rs
+++ b/src/objects/string.rs
@@ -159,7 +159,7 @@ impl<'a> PyStringData<'a> {
                 )),
             },
             PyStringData::Latin1(data) => {
-                if data.iter().all(|&b| b.is_ascii()) {
+                if data.is_ascii() {
                     Ok(Cow::Borrowed(unsafe { str::from_utf8_unchecked(data) }))
                 } else {
                     Ok(Cow::Owned(data.iter().map(|&b| b as char).collect()))
@@ -214,7 +214,7 @@ impl<'a> PyStringData<'a> {
         match self {
             PyStringData::Utf8(data) => String::from_utf8_lossy(data),
             PyStringData::Latin1(data) => {
-                if data.iter().all(|&b| b.is_ascii()) {
+                if data.is_ascii() {
                     Cow::Borrowed(unsafe { str::from_utf8_unchecked(data) })
                 } else {
                     Cow::Owned(data.iter().map(|&b| b as char).collect())
@@ -598,6 +598,7 @@ mod test {
         let py = gil.python();
         let py_string = py.eval("u'x=\\u00e4'", None, None).unwrap();
         let data = py_string.cast_as::<PyString>(py).unwrap().data(py);
+        #[cfg(feature = "python3-sys")]
         if let PyStringData::Latin1(s) = data {
             assert_eq!([b'x', b'=', 0xe4], *s);
         } else {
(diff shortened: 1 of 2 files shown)


src/objects/string.rs
```diff
@@ -583,4 +610,47 @@ mod test {
...
+
+    #[test]
+    fn test_extract_umlaut() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let py_string = py.eval("u'x=\\u00e4'", None, None).unwrap();
+        let data = py_string.cast_as::<PyString>(py).unwrap().data(py);
+        #[cfg(feature = "python3-sys")]
+        {
+            if let PyStringData::Latin1(s) = data {
```
markbt, line 622: Is it worth adding tests for the other `PyStringData` variants?  I assume non-Latin-1 text (e.g. something in Greek) would be Utf16 and non-BMP text (e.g. something with emojis) would be Utf32.

markbt (approved):
This looks good to me.

PR #248, opened 2021-01-10
Bump from paste 0.1 to 1

decathorpe:
This commit also adds excludes for one more file (Makefile), as that file is not useful in published crates.


markbt:
Thanks!

PR #250, opened 2021-01-15
allow all visibility keywords on classes and functions

tdyas:
Teach the `py_class` macros about allowing arbitrary visibility keywords (e.g., `pub(crate)`) on class definitions and the method definitions within that class definition. This PR also makes the `create_instance` method have the same visibility as the class.

[The `vis` capture type for visibility in macro_rules has been available since Rust v1.30.0.](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1300-2018-10-25)  Given that this crate requires v1.32 or higher, this PR does not break the crate's compatibility guarantee.

Note: The ultimate motivation for this PR is to allow me to split a very complicated use of this crate into multiple modules without having to make many types `pub`.

### DIFF
--- a/src/py_class/py_class_impl2.rs
+++ b/src/py_class/py_class_impl2.rs
@@ -387,7 +387,7 @@ macro_rules! py_class_impl {
         }
         /* impl: */ {
             $($imp)*
-            $crate::py_class_impl_item! { $class, $py, __abs__(&$slf,) $res_type; { $($body)* } [] }
+            $crate::py_class_impl_item! { $class, $py, pub, __abs__(&$slf,) $res_type; { $($body)* } [] }
         }
         $members $props
     }};
@@ -417,7 +417,7 @@ macro_rules! py_class_impl {
         }
         /* impl: */ {
             $($imp)*
-            $crate::py_class_impl_item! { $class, $py, __add__() $res_type; { $($body)* } [ { $left : &$crate::PyObject = {} } { $right : &$crate::PyObject = {} } ] }
+            $crate::py_class_impl_item! { $class, $py, pub, __add__() $res_type; { $($body)* } [ { $left : &$crate::PyObject = {} } { $right : &$crate::PyObject = {} } ] }
         }
         $members $props
     }};
@@ -459,7 +459,7 @@ macro_rules! py_class_impl {
         }
         /* impl: */ {
             $($imp)*
-            $crate::py_class_impl_item! { $class, $py, __and__() $res_type; { $($body)* } [ { $left : &$crate::PyObject = {} } { $right : &$crate::PyObject = {} } ] }
+            $crate::py_class_impl_item! { $class, $py, pub, __and__() $res_type; { $($body)* } [ { $left : &$crate::PyObject = {} } { $right : &$crate::PyObject = {} } ] }
         }
         $members $props
     }};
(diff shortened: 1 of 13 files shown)


tdyas:
The CI failures are confusing since I'm not sure they result from this PR:

```
     Running target/debug/deps/test_thread_exit-591aa67f956d4001

running 1 test

test test_thread_exit ... FATAL: exception not rethrown

error: test failed, to rerun pass '--test test_thread_exit'

Caused by:

  process didn't exit successfully: `/home/travis/build/dgrunwald/rust-cpython/target/debug/deps/test_thread_exit-591aa67f956d4001` (signal: 6, SIGABRT: process abort signal)

Makefile:50: recipe for target 'test' failed

make: *** [test] Error 101

The command "make test extensions" exited with 2.
```

Another failure is:

```
error[E0560]: struct `python27_sys::object::PyNumberMethods` has no field named `nb_bool`
   --> tests/test_class.rs:481:1
    |
481 | / py_class!(class Comparisons |py| {
482 | |     data val: i32;
483 | |
484 | |     def __hash__(&self) -> PyResult<i32> {
...   |
490 | |     }
491 | | });
    | |___^ `python27_sys::object::PyNumberMethods` does not have this field
    |
    = note: available fields are: `nb_add`, `nb_subtract`, `nb_multiply`, `nb_divide`, `nb_remainder` ... and 34 others
    = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
error: aborting due to previous error
```

`test_thread_exit` passes locally for me though when I run `cargo test`. Not sure why this is failing in CI.

markbt:
Hi @tdyas, thanks for the contribution!

This looks good to me.  I'm not sure why the tests are failing in CI, let me see if I can understand what's going on.

src/py_class/py_class_impl2.rs
```diff
@@ -487,21 +487,21 @@ macro_rules! py_class_impl {
             $type_slots
             /* as_number */ [
                 $( $nb_slot_name : $nb_slot_value, )*
-                nb_nonzero: $crate::py_class_unary_slot!($class::__bool__, $crate::_detail::libc::c_int, $crate::py_class::slots::BoolConverter),
+                nb_bool: $crate::py_class_unary_slot!($class::__bool__, $crate::_detail::libc::c_int, $crate::py_class::slots::BoolConverter),
```
markbt, line 490: This is the source of at least one error.  The `py_class_impl2.rs` script needs to be run with env `PY=2` in order to generate correct output.  (You should get this if you use `make` to generate them.  Did you run the script manually?)

tdyas: I did. It wasn't obvious to me how to regenerate the files so I ran the script manually. I'll regenerate with `make` and then adds some docs.

tdyas: Okay, pushed a fix with those files regenerated via the Makefile.

markbt: Thanks!

markbt:
`test_thread_exit` looks like it is flaky.  I don't think it tests the right thing, as adjusting the timing slightly makes it fail more often or less often on my machine.  I'll make a separate issue about that.

tdyas:
Pushed some commits to update the README to mention the `Makefile` and to update the `py_class` macro rustdoc to mention the visibility keywords.

markbt (approved):
This all looks good to me.  Will give others a few days to comment in case I have missed anything.

tdyas:
> This all looks good to me. Will give others a few days to comment in case I have missed anything.

Thanks!

tdyas:
Regarding the `create_instance` function, the PR currently makes its visibility follow the class visibility. Would it be better to have its visibility be separately specified? (For example, use a visibility keyword on the `__new__` function for `create_instance`.)

markbt:
> Regarding the `create_instance` function, the PR currently makes its visibility follow the class visibility.

I think this is fine for now.  If we find a need for more granular visibility, we can add it later as a keyword to `__new__` as you suggest.

tdyas:
> I think this is fine for now. If we find a need for more granular visibility, we can add it later as a keyword to __new__ as you suggest.

As this PR is currently structured, `__new__` can now take a visibility keyword which is applied to generated `__new__` method. Should visibility be disallowed on `__new__` for now then?

markbt:
> As this PR is currently structured, `__new__` can now take a visibility keyword which is applied to generated `__new__` method. Should visibility be disallowed on `__new__` for now then?

I think it's fine as it is.   I doubt we're going to need to change the visibility of `create_instance` on the Rust side.

tdyas:
> I think it's fine as it is. I doubt we're going to need to change the visibility of `create_instance` on the Rust side.

Sounds good.

tdyas:
Rebased due to merge conflict in README.md.

tdyas:
Just wanted to follow up on getting this PR merged.

markbt:
Thanks!

PR #251, opened 2021-01-15
implement ToPyObject for unit value ()

tdyas:
Implement ToPyObject for the unit value (i.e., `()`). This provides a more ergonomic way to declare functions that return `None` since this conversion allows the return value in the Rust declaration to be `PyResult<()>` instead of `PyResult<PyObject>` plus also having to call `Ok(py.None())`.

### DIFF
--- a/src/objects/mod.rs
+++ b/src/objects/mod.rs
@@ -146,6 +146,7 @@ mod set;
 mod string;
 mod tuple;
 mod typeobject;
+mod unit;
 
 #[cfg(feature = "python27-sys")]
 pub mod oldstyle;
--- a/src/objects/unit.rs
+++ b/src/objects/unit.rs
@@ -1,0 +1,29 @@
+use super::PyObject;
+use crate::conversion::ToPyObject;
+use crate::err::PyResult;
+use crate::ffi;
+use crate::python::Python;
+
+impl ToPyObject for () {
+    type ObjectType = PyObject;
+
+    #[inline]
+    fn to_py_object(&self, py: Python) -> Self::ObjectType {
+        py.None()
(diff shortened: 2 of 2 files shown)


tdyas:
Should this also override `into_py_object` and/or `with_borrowed_ptr`?

markbt:
The reason we have not implemented this before is that `()` is also the empty tuple, and it's ambiguous as to whether `()` should be converted to `None` (for void function return) or `tuple()` (for empty function arguments).

What I've used successfully elsewhere is a `PyNone` sentinel type that can be used in function signatures (i.e. `PyResult<PyNone>` and `Ok(PyNone)`).  You can see the type defined here: https://github.com/facebookexperimental/eden/blob/master/eden/scm/lib/cpython-ext/src/none.rs

If there's more interest in that, I could see about contributing it upstream.

tdyas:
I noticed that this crate has [a `NoArgs` sentinnel type](https://github.com/dgrunwald/rust-cpython/blob/633dafc2c9e0ec22f8ff2479dd1cdadeb4ac0f40/src/objects/tuple.rs#L200-L235) to represent the empty tuple for function arguments. Does that resolve the ambiguity?

markbt:
Only on one direction.  `()` becoming `None` might be surprising for someone who was expecting the empty tuple.  It's better if it doesn't convert at all, with a sentinel for both.  I'm going to submit a PR for upstreaming my `PyNone` sentinel.

markbt:
I've sent #253 which upstreams my `PyNone` implementation.

tdyas:
Makes sense. Will close this PR then.

PR #252, opened 2021-01-15
Disable test_thread_exit on Rust >= 1.48

markbt:
This test no longer works, so disable it.
I'm not sure what to do about the wider question of correctly handling `pthread_exit`.


quark-zju:
Thanks!

PR #255, opened 2021-01-24
Increase MSRV to 1.41.1

markbt:
The `thread_local` dependency now requires a minimum Rust version of
1.36.  Increase our MSRV to 1.41.1, which is the version in Debian
stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,14 +8,14 @@ python:
   - "3.8"
   - "3.9"
 env:
-  - RUST_VERSION=1.32.0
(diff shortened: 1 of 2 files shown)


markbt:
Tagging @gracinet as I think Mercurial wanting to support Debian stable was the reasoning behind the previous MSRV.

gracinet:
@markbt I can confirm yes, though these days @Alphare  is much more active than me on the Rust Mercurial front

Alphare:
I can confirm that this bump is fine for Mercurial. :)
Thanks for checking

PR #256, opened 2021-01-24
GitHub actions

markbt:
As travis-ci.org is shutting down, and CI builds for open source are going to be more limited, let's also add Github actions, so we can move away from Travis later on if we wish.

The Unix tests should be the same as the Travis ones.  For Windows, I don't test the extensions as that doesn't seem to work, the same as in the Appveyor definition.

To keep the matrix manageable, I've limited the number of Python versions.  We can increase if we wish, although I noticed that the Python 3.4 build didn't work when I tested it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,76 @@
+name: Build and Test
+
+on:
+  push:
+    branches: [ master, github-actions ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


markbt:
See https://github.com/markbt/rust-cpython/actions/runs/507886337 for an example run.

Issue #257, opened 2021-02-26
PyObjectArenaAllocator error


ryancinsight:
Hello,

On pythons website:
Pyobjectarenaallocator is denoted as:
void *ctx | user context passed as first argument
void* alloc(void *ctx, size_t size) | allocate an arena of size bytes
void free(void *ctx, size_t size, void *ptr) | free an arena

yet you have written as:
pub struct PyObjectArenaAllocator {
    pub ctx: *mut c_void,
    pub alloc: Option<extern "C" fn(ctx: *mut c_void, size: size_t) -> *mut c_void>,
    pub free: Option<extern "C" fn(ctx: *mut c_void, ptr: *mut c_void, size: size_t) -> ()>,

switching the size and pointer in free...

Wouldn't this cause memory allocation issues?

markbt:
As discussed in #258 this is actually a bug in the Python docs.  The `rust-cpython` code matches the CPython header.

PR #258, opened 2021-02-26
Correct PyObjectArenaAllocator to match description in python docs

ryancinsight:
Pyobjectarenaallocator free should be [free(void *ctx, size_t size, void *ptr)]

according to https://docs.python.org/3/c-api/memory.html

#257 
@indygreg

### DIFF
--- a/python3-sys/src/objimpl.rs
+++ b/python3-sys/src/objimpl.rs
@@ -48,7 +48,7 @@ pub use self::_PyObject_DebugRealloc as PyObject_Realloc;
 pub struct PyObjectArenaAllocator {
     pub ctx: *mut c_void,
     pub alloc: Option<extern "C" fn(ctx: *mut c_void, size: size_t) -> *mut c_void>,
-    pub free: Option<extern "C" fn(ctx: *mut c_void, ptr: *mut c_void, size: size_t) -> ()>,
+    pub free: Option<extern "C" fn(ctx: *mut c_void, size: size_t, ptr: *mut c_void) -> ()>,
 }
 #[cfg(all(not(Py_LIMITED_API), Py_3_4))]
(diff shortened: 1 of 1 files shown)


indygreg (approved):
This change looks good to me!

But I'm not a project maintainer so I can't merge it for you. I'm also unsure if this should be documented in release notes anywhere. I'll let an actual maintainer chime in :)

markbt:
Thanks!
It looks like this definition isn't used elsewhere in rust-cpython, so it would only affect anyone accessing it directly through `python3-sys`.  I'm guessing not many people have done that.

markbt:
Oh, I double-checked the code, and it looks like it's actually the **Python docs** that are wrong: [see the definition in CPython itself](https://github.com/python/cpython/blob/master/Include/cpython/objimpl.h#L72).

ryancinsight:
Yeah, we are trying to customize the Python allocators used in portable Python packages. Ahh, I should have tracked it down further, I’ll get in touch with them to figure this out.

indygreg:
> Oh, I double-checked the code, and it looks like it's actually the **Python docs** that are wrong: [see the definition in CPython itself](https://github.com/python/cpython/blob/master/Include/cpython/objimpl.h#L72).

CC @vstinner

markbt:
Going to close this PR as it's the Python docs that are wrong, and I don't want it to get accidentally merged.

vstinner:
> it's the Python docs that are wrong

Oops. I proposed to https://github.com/python/cpython/pull/24795 to fix.

The PEP is as old as the PEP 445 implementation: https://github.com/python/cpython/commit/0507bf56f0b1064a179464eb3a9d524fd6dad815

PR #259, opened 2021-03-05
Fix or suppress warnings and format code

markbt:
Fix or suppress all warnings, including clippy lints.  A new `make clippy` target is added to run clippy.

Reformat using the latest versions of `rustfmt`.

There were some formatting conflicts caused by the commented out prototypes in the `sys` crates.  I have switched these to using an `ignore!` macro so that they can be formatted better, and also makes it clearer what's happening.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,9 @@ endif
 doc: build
 	cargo doc --no-deps $(CARGO_FLAGS)
 
+clippy:
+	cargo clippy $(CARGO_FLAGS)
+
 extensions: build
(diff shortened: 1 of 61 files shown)


markbt:
With this in place, the only warning is in `nightly` where `specialization` warns that it is incomplete.  Unfortunately `min_specialization` doesn't provide what we need, so we will have to live with that.

If there are no objections, I will merge this and my other PRs and make a 0.6 release soon.

dgrunwald:
I had a cursory look at your PRs and have no objections.

PR #261, opened 2021-04-13
Build scripts should invalidate when relevant env vars change

stuhood:
When changing environment variables relevant to the build, Cargo should re-run our build scripts.

Tell cargo [that we have accessed env vars](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script).


markbt:
Thanks!

PR #264, opened 2021-07-08
Include Rust panic messages in Python exception

SimonSapin:
When Python calls a Rust function, an unhandled Rust panic is turned into a Python `SystemError` exception. Previously, the value of that exception was just "Rust panic" without more details.

This adds the panic message to that value.


SimonSapin:
This does not (yet, hopefully) work as I expected. Even though https://github.com/rust-lang/rust/blob/1.53.0/library/std/src/panicking.rs#L452-L472 looks like the concrete type behind `dyn Any` should be `String`, when I run this test `downcast_ref::<String>()` returns `None`.

SimonSapin:
Thanks to [Gus Wynn on Rust’s Zulip](https://zulip-archive.rust-lang.org/219381tlibs/07261PanicmessagefromcatchunwindsdynAny.html#245343529) who pointed out the problem in my previous version of this PR: there was two layers of `dyn Any` vtables. See code comments in the amended commit.

markbt (approved):
Nice!  Thanks!

markbt:
The AppVeyor failure seems to be spurious: it's having problems with the network downloading crates.  Does anyone know if they've added some kind of network sandboxing recently?

SimonSapin:
I have no idea. I haven’t used AppVeyor in a long time.

SimonSapin:
@markbt What can be done do unblock this?

markbt:
The code looks fine, and the AppVeyor failures look unrelated, so I'll just merge it.

PR #266, opened 2021-07-31
Expose python3.9 feature.

DimitrisJim:
Unless I'm missing something, this seems to have been omitted in #243


markbt:
It does indeed look like it was just missed.  Thanks!

PR #267, opened 2021-08-09
Add support for the number protocol

Techcable:
NOTE: We now get warnings because of the modularization of the macro system
specified in RFC rust-lang/rfcs#1561

However, fixing these warnings is out of scope for this PR.

### DIFF
--- a/src/objectprotocol/number.rs
+++ b/src/objectprotocol/number.rs
@@ -1,0 +1,323 @@
+use std::cmp::Ordering;
+use std::fmt;
+
+use crate::conversion::ToPyObject;
+use crate::err::{self, PyErr, PyResult};
+use crate::ffi;
+use crate::objects::{PyObject, PyInt, PyLong, PyFloat};
+use crate::python::{Python, PythonObject, ToPythonPointer};
+
+
+use super::ObjectProtocol;
+
+/// Operations on numeric objects
+pub trait NumberProtocol: ObjectProtocol {
+    /// Perform addition (self + other)
+    ///
+    /// Invokes the `__add__` magic-method
+    #[inline]
+    fn add(&self, py: Python, other: impl ToPyObject) -> PyResult<PyObject> {
+        other.with_borrowed_ptr(py, |other| unsafe {
+            err::result_from_owned_ptr(py, ffi::PyNumber_Add(self.as_ptr(), other))
+        })
+    }
+    /// Perform subtraction (self - other)
+    ///
+    /// Invokes the `__sub__` magic-method
+    #[inline]
+    fn subtract(&self, py: Python, other: impl ToPyObject) -> PyResult<PyObject> {
+        other.with_borrowed_ptr(py, |other| unsafe {
+            err::result_from_owned_ptr(py, ffi::PyNumber_Subtract(self.as_ptr(), other))
+        })
+    }
+    /// Perform multiplication (self * other)
+    ///
+    /// Invokes the `__mul__` magic-method
+    #[inline]
+    fn multiply(&self, py: Python, other: impl ToPyObject) -> PyResult<PyObject> {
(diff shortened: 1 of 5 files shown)


src/objectprotocol/number.rs
```diff
@@ -0,0 +1,323 @@
...
+    /// equivalent to the Python expression `self // other`,
+    /// 
+    /// This method was added in Python 3.
+    /// If compiling against Python 2, it unconditional throws an error.
+    ///
+    /// Invokes the `__floordiv__` magic-method.
+    #[inline]
+    fn div_floor(&self, py: Python, other: impl ToPyObject) -> PyResult<PyObject> {       
+        other.with_borrowed_ptr(py, |other| unsafe {
+            err::result_from_owned_ptr(py, ffi::PyNumber_TrueDivide(self.as_ptr(), other))
```
markbt, line 115: Should this be calling `PyNumber_FloorDivide`?

src/objectprotocol/number.rs
```diff
@@ -0,0 +1,323 @@
...
+    /// The presence of this method indicates 
+    /// this object is an integer type.
+    ///
+    /// Calls the `__index__` magic-method.
+    ///
+    /// See also: [Documentation on the corresponding magic-method](https://docs.python.org/3/reference/datamodel.html?highlight=__index__#object.__index__)
+    #[inline]
+    fn as_int_index(&self, py: Python) -> PyResult<PyLong> {
+        let obj = unsafe {
+            err::result_from_owned_ptr(py, ffi::PyNumber_Long(self.as_ptr()))?
```
markbt, line 265: And `PyNumber_Index` here?

markbt (changes requested):
Generally looks good, thanks!

A couple of them look like they're calling the wrong CPython function.  Can you check and fix these?

src/objectprotocol/number.rs
Techcable: Fixed. Thank you!

Techcable:
I fixed the two mistakes you noticed and changed the naming to be more consistent with the corresponding magic methods and CAPI calls.

src/objectprotocol/number.rs
```diff
@@ -0,0 +1,323 @@
...
+    ///
+    /// See [PEP 0456](https://www.python.org/dev/peps/pep-0465/) for details.
+    #[inline]
+    fn matrix_multiply(&self, py: Python, other: impl ToPyObject) -> PyResult<PyObject> {
+        #[cfg(not(any(feature = "python3-4", feature = "python2-sys")))] {
+            other.with_borrowed_ptr(py, |other| unsafe {
+                err::result_from_owned_ptr(py, ffi::PyNumber_MatrixMultiply(self.as_ptr(), other))
+            })
+        }
+        #[cfg(any(feature = "python3-4", feature = "python2-sys"))] {
```
dgrunwald, line 57: This feature checks if the user explicitly requested python3-4 as a cargo feature.
It won't be set when python 3.4 is instead selected via `PYTHON_SYS_EXECUTABLE`.

Fixing this will require solving https://github.com/dgrunwald/rust-cpython/pull/269#discussion_r696136064 in some way.

src/objectprotocol/number.rs
```diff
@@ -0,0 +1,323 @@
...
+    }
+    /// Convert this object to an integer,
+    /// equivalent to the builtin function `int(self)`
+    ///
+    /// Invokes the `__int__` magic-method.
+    ///
+    /// Throws an exception if unable to perform
+    /// the conversion.
+    #[inline]
+    fn as_int(&self, py: Python) -> PyResult<PyLong> {
```
dgrunwald, line 233: Following the Rust naming conventions, `to_int` would be a more appropriate name -- to me, `as_` implies the function returns some form of reference.

Issue #263, opened 2021-07-06
Use GILProtected for capsule caching?


SimonSapin:
The `py_capsule` and `py_capsule_fn` macros generate code that "cache" the pointer obtained from a capsule in a `static mut` item, with a `std::sync::Once` in a second `static` item for thread-safety.

Although the synchronization cost of `Once` is small, it seems unnecessary here since code touching this is always holding the GIL. Could this cache be implemented with a single `static` item of type `cpython::GILProtected<std::cell::Cell<_>>` instead?

markbt:
This is probably a question for @gracinet.

SimonSapin:
I’ve submitted https://github.com/dgrunwald/rust-cpython/pull/268. `Cell` turned out not to be enough, this uses `OnceCell` instead. The fact that no cache-related `unsafe` code is needed anymore reinforces my belief that this approach is sound.

PR #268, opened 2021-08-12
Cheaper and safe caching for py_capsule! and py_capsule_fn!

SimonSapin:
The `py_capsule!` and `py_capsule_fn!` macros generate `retrieve` functions that cache their results for subsequent calls.

Prior to this commit, caching is done with a generated unsafe `static mut` item whose access is made thread-safe by a `std::sync::Once` on the side. However this synchronization is unnecessary since `retreive` takes a `Python<'_>` parameter that indicates that the GIL is held.

This changes the cache to use safe `static` item instead, with `GILProtected` for synchronization-free thread-safety and `OnceCell` for interior mutability. As a bonus, this removes the need for cache-related `unsafe` code in generated `retrieve` functions.

This adds a dependency to the `once_cell` crate, which can be removed when `OnceCell` becomes stable in the standard library: https://github.com/rust-lang/rust/issues/74465

Alternatively `OnceCell` could be replaced with `UnsafeCell` plus some `unsafe` code, effectively inlining the core of the logic of `OnceCell`. `RefCell` might work too.

Fixes https://github.com/dgrunwald/rust-cpython/issues/263


SimonSapin:
@markbt, any thought on this?

PR #272, opened 2021-09-15
Add missing PyNumeric class methods.

fsh:
I hit a wall in my own project because many of the numeric methods (truediv, floordiv, mod, pow, etc.) were set to produce a hard error, so I added them here.

I also added base documentation and a test for them.

I am not familiar at all with the code of this project, so if I did something wrong, let me know.

### DIFF
--- a/src/py_class/py_class_impl.py
+++ b/src/py_class/py_class_impl.py
@@ -739,15 +739,23 @@ def call_operator(special_name, slot):
         value_args='$class::%s' % special_name)
 
 @special_method
-def binary_numeric_operator(special_name, slot):
+def numeric_operator(special_name, arity, slot):
+    if arity == 'binary':
+      argnames = ['left', 'right']
+    elif arity == 'ternary':
+      argnames = ['left', 'right', 'ex']
+    else:
+      error("Invalid arity %s" % arity)
+    def_args_str = ', '.join("$%s:ident" % name for name in argnames)
+    impl_args_str = ' '.join("{ $%s : &$crate::PyObject = {} }" % name for name in argnames)
     generate_case(
-        pattern='def %s($left:ident, $right:ident) -> $res_type:ty { $($body:tt)* }'
-            % special_name,
-        new_impl='$crate::py_class_impl_item! { $class, $py, pub, %s() $res_type; { $($body)* } ' % special_name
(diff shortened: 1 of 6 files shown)


dgrunwald:
The tests fail to compile with Python 2.7.

This will need some `#[cfg(Py_3_5)]`, which in turn requires solving the mystery of why this cfg doesn't seem to work in the cpython crate, even though the [build.rs](https://github.com/dgrunwald/rust-cpython/blob/master/build.rs#L45-L46) should be copying over the compilation flags from the python3-sys crate.

fsh:
Oof, Python 2... OK.

Well, I'll just pull out the matmul in PY2 case. As for the tests I can just separate them.

It also seems I also have to pull floor_divide and true_divide from Python 2 compiles, because they require `Py_TPFLAGS_HAVE_CLASS` flag to be set on type object... And I'm scared of adding it by default for no reason -- especially since I personally don't care about Python 2.

markbt:
> Oof, Python 2... OK.

I think it's fine to do the bare minimum for Python 2, so what you have done is great.  Thank you.

There's a separate question of how much longer we support Python 2 for.  FWIW we're now migrated full to Python 3.  Another hold-out was Mercurial - @gracinet or @Alphare might know what the current state is there.

Alphare:
Thanks for the ping!
Next release cycle will remove Python 2 support from Mercurial, so normally starting 1st of November for the 6.1 cycle.

src/py_class/py_class_impl.py
```diff
@@ -877,7 +885,7 @@ def inplace_numeric_operator(special_name, slot):
     '__itruediv__': inplace_numeric_operator('nb_inplace_true_divide'),
     '__ifloordiv__': inplace_numeric_operator('nb_inplace_floor_divide'),
     '__imod__': inplace_numeric_operator('nb_inplace_remainder'),
-    '__ipow__': unimplemented(),
+    '__ipow__': inplace_numeric_operator('nb_inplace_power'),
```
dgrunwald, line 888: `nb_inplace_power` is a `ternaryfunc` unlike the other in-place operators.
So I think this won't work as-is (or at a minimum, needs a test case).

dgrunwald:
I've merged this PR but reverted the `__ipow__` change.

Thanks for your contribution!

PR #283, opened 2022-07-22
The scope of the unsafe block can be appropriately reduced

cactter:


dgrunwald:
It doesn't make sense to reduce the size of these unsafe blocks -- the code that you moved outside the blocks is still relevant to their safety and needs to be reviewed just as carefully as the unsafe code itself.

PR #279, opened 2022-05-14
Initial support for Python 3.11 beta

Techcable:
This is obviously experimental but it works okay.

All the tests pass and the example prints the following:

> $ cargo run --example hello --features python-3-11
> Hello nicholas, I'm Python 3.11.0b1 (v3.11.0b1:8d32a5c8c4, May  6 2022, 22:45:29) [Clang 13.0.0 (clang-1300.0.29.30)]

I figure it's best to be proactive with these change, so that way we are ready as soon as 3.11 proper is released (October I think).

MAin difference in 3.11 that frame object is now private (and lazy loaded).
This results in a performance improvement but makes it harder for debuggers.
Also the internal ordering of code objects has completely changed as well.

PR #284, opened 2022-09-19
Replace -Z ast-json in the check for undefined symbols

tschorr:
`-Z ast-json` ist no longer available in Rust nightly, see also https://github.com/dgrunwald/rust-cpython/pull/279#issuecomment-1159658662. Suggestion: Let `nm` report undefined symbols. This should also work for Rust stable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,10 @@ build: src/py_class/py_class_impl2.rs src/py_class/py_class_impl3.rs python27-sy
 
 test: build
 	cargo test $(CARGO_FLAGS)
+ifeq ($(NIGHTLY),1)
+# unpretty=ast-tree,expanded output is only supported on nightly
 	python$(PY) tests/check_symbols.py
+endif
 
 doc: build
 	cargo doc --no-deps $(CARGO_FLAGS)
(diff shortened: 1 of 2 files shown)


dgrunwald:
I think `nm` will only report functions that are actually called, I wanted to check all functions that are declared even if unused.

tschorr:
> all functions that are declared even if unused.

I'm not exactly sure what that should cover and I don't have the `-Z ast-json` output to compare the results. Could you elaborate?
`nm -u` will report undefined symbols (those external to each object file). It seems to make sense to check if those are provided in the `so_files`. But there is also `nm --defined-only` that will report symbols in the text and data sections.

There is also the `nightly -Z unpretty=ast-tree,expanded` flag to `rustc` that might provide what you're looking for. But the output ist different from `-Z ast-json` (e.g. it's not json).

tschorr:
After looking a bit more into the output of the different commands, I have decided to go with `-Z unpretty=ast-tree,expanded` as it  seems closer to the original test than the other options. Also, the `.rlib` contents differ a lot depending on the link mode.

PR #285, opened 2022-10-08
Update actions, try other action for providing the rust toolchain

tschorr:
https://github.com/dtolnay/rust-toolchain seems to be active, maybe this can also fix the build problems on the py3.11 branch.


dgrunwald:
Thanks! That solved the build problem on py3.11.

Issue #213, opened 2020-03-30
Linking against pythonXX.lib fails on windows within virtualenv


striggldy:
When a virtualenv (created with the venv or virtualenv module, pyvenv not tested) is active, building on windows fails with `fatal error LNK1181`, because the pythonXX.lib cannot be found.

In
https://github.com/dgrunwald/rust-cpython/blob/387e87d9deb6b678508888239f9f87dc36973d3f/python3-sys/build.rs#L367-L368

the rustc lib search path `{exec_prefix}/libs` will be added, which is valid for a system interpreter, but not for virtualenvs created with the modules mentioned above. There's no `LIBDIR` set as well when running `python -c "import sysconfig;print(sysconfig.get_config_var('LIBDIR')"`

Is there a way to specifiy additional link search paths for rust_cpython?

nemjit001:
Hey, I found a fix for this. Seems like checking the ```sys.exec_prefix``` instead of ```sys.base_prefix``` in the ```get_config_from_interpreter``` function causes the issue.

I got it working locally. I'll open a PR and see if I can get it approved.

PR #288, opened 2022-12-01
fixed linking python when venv is active

nemjit001:
By checking for base_prefix instead of exec_prefix, linking with pythonXY.lib works even when a virtual environment is active.

Changes:
  - printing sys.base_prefix instead of sys.exec_prefix

This PR references issue #213


nemjit001:
It seems that the failing jobs have something to do with macOS not wanting to compile due to linking errors in dependencies, but the windows tests are passing.

Is this a known issue with the tests?

PR #289, opened 2023-01-04
Fix build status badge

tschorr:
Editing while reading - the status currently shown is from Travis CI and 2 years old.


tschorr:
Tests are failing for `macos-latest` and Rust 1.41.1. Maybe this is a problem with `dtolnay/rust-toolchain` and that specific Rust version on MacOS.
In any case the failure is not related to this PR (nor to the changes in https://github.com/dgrunwald/rust-cpython/pull/288).

PR #299, opened 2025-02-12
Fix a clippy warning

zzl0:
This diff fixes the clippy warning about manual_strip: https://rust-lang.github.io/rust-clippy/master/index.html#manual_strip

Screenshot 2025-02-12 at 10 36 06 AM


markbt:
`String::strip_prefix` was added in 1.45 but our MSRV is 1.41.  1.45 is nearly 5 years old at this point, so perhaps we could consider bumping the MSRV.  I think the reason for holding back was the version of Rust in Debian stable, so will need to check that.

